PR #139, opened 2020-11-03
[require-id-when-available] fix for inline fragments on interfaces

joual:
With a schema like

```graphql
type Query {
  vehicles: [Vehicle!]!
}

interface Vehicle {
  id: ID!
}

type Car implements Vehicle {
  id: ID!
  mileage: Int
}
```

and a query like

```graphql
query { 
  vehicles { 
    id 
    ...on Car { 
      mileage 
    } 
  } 
}
```

I currently get a lint error because it'll say that `id` should also be present in the Car selection set.

This PR attempts to fix this issue. I had a problem with the types on `node.parent` which is why I typed it as `any`. Not quite sure what I should type it as to get it to work.

### DIFF
--- a/packages/plugin/src/rules/require-id-when-available.ts
+++ b/packages/plugin/src/rules/require-id-when-available.ts
@@ -67,7 +67,7 @@ const rule: GraphQLESLintRule<RequireIdWhenAvailableRuleConfig, true> = {
               //   parent.parent.kind === 'SelectionSet' &&
               //   !!parent.parent.selections.find(s => s.kind === 'Field' && s.name.value === fieldName);
 
-              if (!hasIdFieldInSelectionSet && !hasIdFieldInInterfaceSelectionSet) {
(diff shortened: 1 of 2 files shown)


dotansimha:
Great catch, thank you @joual !

joual:
Thank you! I'll be watching for the release with the fix. I'm currently trying to integrate it into our tooling over at SurveyMonkey and so far it's the blocking issue. Cheers 🍻

dotansimha:
Cool, thank you @joual . I'll cut a release today or tomorrow :D

joual:
Hey @dotansimha, 👋  Just wanted to send a ping to remind you about the release just in case it slipped your mind. :) If you remembered but still have other things to include that's totally fine!

dotansimha:
Oops, done :) `@graphql-eslint/eslint-plugin@0.4.2` @joual thanks!

Issue #25, opened 2020-09-26
Allow to access all loaded operation files 


dotansimha:
Some rules might needs access to other loaded operations - maybe we can have a util on the context level for getting a fragment by name or something like that.

This might be done using ESLint as core, since we need to make sure to load the same files ESLint loaded (as defined in the configuration - `exclude`, `extension` and so on)

ilyavolodin:
Unfortunately, this isn't really possible to do in ESLint, since ESLint lints a file as they load. I.e. even if another file was loaded and linted before, you have no access to its content. ESLint also doesn't guarantee an order of linting, so you can't predict which files will be loaded before the current file.
eslint-plugin-import solved this by basically have a singleton instance that's called from every rule. It will load scan all your dependencies once, and every consecutive request will be served from cache.

dotansimha:
I was thinking to expose a function in the parserServices that will use graphql-tools to load siblings files

ilyavolodin:
Yes, that's possible to do as parserService, but then the parser will always require graphql-config file for any linting. I would love to avoid having another file required for all rules to function.

dotansimha:
@ilyavolodin I was thinking to do that lazily, so the parserService will expose a function the loads (and caches) the graphql files. So it will get loaded only with specific rules, and only if the relevant AST was found, and it will be cached.

ilyavolodin:
But since the order of linting is not guaranteed, you never know when is the last call to the `parserService` is going to occur. More than that, if you lazily cache files, when you lint second file (out of, let's say, 20), you will very incomplete schema.
In order to have predictable caching, linting would have to happen in two stages, and ESLint only does one stage linting. I keep wanting to propose an RFC that would add 2 new hooks for plugins for ESLint, one that happens before any linting begins, and one that happens at the end, once everything is done, but before reports are generated.

coreylight:
I think my project requires this to be able to do any valid linting.
We put our fragments into separate files that get picked up by [graphql codegen](https://graphql-code-generator.com/) correctly. So we'd have to tell the parser to look at those somehow.

dotansimha:
Yeah, @coreylight , that was the exact inspiration for that. 
@ilyavolodin you are right about ESLint loading and execution process, but I don't aim to use ESlint for that. 
[...]

PR #142, opened 2020-11-04
Improve performance for parserServices, and allow access to sibling operations in rules

dotansimha:
Related: https://github.com/dotansimha/graphql-eslint/issues/25

### DIFF
--- a/packages/plugin/tests/require-id-when-available.spec.ts
+++ b/packages/plugin/tests/require-id-when-available.spec.ts
@@ -1,5 +1,6 @@
 import { GraphQLRuleTester } from '../src/testkit';
 import rule from '../src/rules/require-id-when-available';
+import { ParserOptions } from '../src/types';
 
 const TEST_SCHEMA = /* GraphQL */ `
   type Query {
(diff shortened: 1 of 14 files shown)


packages/plugin/tests/require-id-when-available.spec.ts
```diff
@@ -53,6 +65,11 @@ ruleTester.runGraphQLTests('require-id-when-available', rule, {
     },
   ],
   invalid: [
+    {
```
henryqdineen, line 68: I'm not sure if this one should be invalid. Was there some defect with the previous solution that I'm not following. Thanks!

dotansimha: Should be moved to `valid`, you are right. 
The thing here: we need to test both edges of usage: consumed clients, and parent selection sets (in case of inline fragments). I will fix that soon, thank you!

dotansimha: Fixed and available in `@graphql-eslint/eslint-plugin@0.5.0`. Thanks @henryqdineen !

Issue #75, opened 2020-10-05
limit input-name to Mutation fields only


BigsonLvrocha:
There is conflict with the rule input-name where it will return error on connection fields, node id input etc.

dotansimha:
Hi @BigsonLvrocha ! Can you please share an example? 
fyi @ilyavolodin

BigsonLvrocha:
This is the base query for a schema following relay's specification
![Screenshot from 2020-10-05 16-28-01](https://user-images.githubusercontent.com/14062008/95123393-443f0880-0728-11eb-9cd6-1c58dd7dfbfa.png)

ilyavolodin:
Oh, good find. Yes, that seems like a bug

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@0.4.3`, thank you @henryqdineen !

PR #156, opened 2020-11-18
input-name: Ignore non-Mutation fields.

henryqdineen:
Related https://github.com/dotansimha/graphql-eslint/issues/75, which we also ran into.

This PR updates the `input-name` rule so it will ignore any fields that aren't on the `Mutation` type. This assumes that the schema is using the default `Mutation` type name. If we want to support a custom mutation type name we could add it as an option or alternatively parse for the `schema { mutation: MyMutationRootType }`

Thanks!

### DIFF
--- a/.changeset/wise-paws-speak.md
+++ b/.changeset/wise-paws-speak.md
@@ -1,0 +1,5 @@
+---
+'@graphql-eslint/eslint-plugin': patch
+---
+
+Fix a bug in the input-name rule to make sure it only checks fields on the Mutation type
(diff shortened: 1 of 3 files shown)


dotansimha:
Awesome catch, thank you @henryqdineen ! 
Can you please add a changeset for that change? (run `yarn changeset` and it will guide you)

henryqdineen:
Thanks @dotansimha. I just pushed the changeset.

dotansimha:
Awesome, thank you. Merging, and will release right after!

PR #165, opened 2020-11-21
docs: fix syntax on require-description example

henryqdineen:
When trying a a copy/paste I noticed a missing `]` this rule configuration. Thanks!


dotansimha:
Awesome! Thank you!

Issue #195, opened 2020-12-05
[naming-convention] Naming convention rule should allow more options  


eddeee888:
Currently, the only thing this rule allows is checking format which is a very good start!

I think it would be even better if we could allow other types of checks such as:

- suffix e.g. Inputs must have `Input` suffix: `CreateMessageInput`
- allows special characters. e.g Fragments must be Pascal case but can have special characters in between: `CreateUser_User`

One way I think we could do it is allowing an object instead of strings for the rule e.g.

```
"@graphql-eslint/naming-convention": [
          "error",
          {
            "InputObjectTypeDefinition": {
               "format": "PascalCase",
               "suffix": "Input"
            },
            "FragmentDefinition": { 
               "format": "PascalCase",
               "allowSpecialCharacters": "_-"
          }
        ]
```

Alternatively, maybe we could allow passing custom regex?

If this makes sense, I can try creating a PR for it :)

dotansimha:
Thank you @eddeee888! 

For inputs name, you can use this rule: https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/input-name.md 

Regarding the rest of the things you mentioned, I think we can add that to the rule itself as well, to allow more flexibility.

PRs are always welcome!

eddeee888:
> you can use this rule: https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/input-name.md

Hi @dotansimha, I tried using the [input name rule]( https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/input-name.md) but it currently has some tight restrictions ( maybe it needs its own issue ):
- It's only being applied to mutations, but users may want to use it for queries as well
- It's forcing the case of the input type e.g.
```graphql
type Mutation {
  postMessage(input: PostMessageInput!): Message! 
}
```
In this example, it fails because it wants the input type to be `postMessageInput` instead of `PostMessageInput`. However, it might conflict with the `naming-convention` rule if we set `InputObjectTypeDefinition` to `PascalCase`


> Regarding the rest of the things you mentioned, I think we can add that to the rule itself to allow more flexibility.

By "add that to the rule itself", do you mean I should update this [file](https://github.com/dotansimha/graphql-eslint/blob/master/packages/plugin/src/rules/naming-convention.ts)?
Maybe an example of how the config would look if it's different from what I posted? :)

eddeee888:
[...]

PR #201, opened 2020-12-07
[naming-convention] Update property schema to accept objects. Add prefix and suffix option

eddeee888:
To make this rule more flexible, it needs to be able to take an object as well as format style string. For example, we might want to check whether a field has a certain suffix. More details: https://github.com/dotansimha/graphql-eslint/issues/195

The config is normalized into an object before passing into the validation function.

Note that we can no longer rely on the default enum check because it also tries to validate that an object is part of the enum. Therefore, the validation function needs to check the format enums now.

Also, added a suffix option. The targeted property must have the correct suffix or it will error.

### TODO

- [x] Update docs ?

### DIFF
--- a/packages/plugin/src/rules/naming-convention.ts
+++ b/packages/plugin/src/rules/naming-convention.ts
@@ -15,16 +15,20 @@ interface CheckNameFormatParams {
   style?: ValidNaming;
   leadingUnderscore: 'allow' | 'forbid';
   trailingUnderscore: 'allow' | 'forbid';
+  prefix: string;
   suffix: string;
 }
 function checkNameFormat(params: CheckNameFormatParams): { ok: false; errorMessage: string } | { ok: true } {
-  const { value, style, leadingUnderscore, trailingUnderscore, suffix } = params;
+  const { value, style, leadingUnderscore, trailingUnderscore, suffix, prefix } = params;
   let name = value;
   if (leadingUnderscore === 'allow') {
     [, name] = name.match(/^_*(.*)$/);
   }
   if (trailingUnderscore === 'allow') {
     name = name.replace(/_*$/, '');
+  }
+  if (prefix && !name.startsWith(prefix)) {
+    return { ok: false, errorMessage: '{{nodeType}} name "{{nodeName}}" should have "{{prefix}}" prefix' };
   }
   if (suffix && !name.endsWith(suffix)) {
     return { ok: false, errorMessage: '{{nodeType}} name "{{nodeName}}" should have "{{suffix}}" suffix' };
@@ -63,6 +67,7 @@ const schemaOption = {
 interface PropertySchema {
   style?: ValidNaming;
   suffix?: string;
+  prefix?: string;
 }
 
 type NamingConventionRuleConfig = [
@@ -150,12 +155,13 @@ const rule: GraphQLESLintRule<NamingConventionRuleConfig> = {
       ...(context.options[0] || {}),
     };
 
-    const checkNode = (node, style, nodeType, suffix = '') => {
+    const checkNode = (node, style, nodeType, suffix = '', prefix = '') => {
       const result = checkNameFormat({
         value: node.value,
         style,
         leadingUnderscore: options.leadingUnderscore,
         trailingUnderscore: options.trailingUnderscore,
+        prefix: prefix,
         suffix: suffix,
       });
       if (result.ok === false) {
@@ -163,6 +169,7 @@ const rule: GraphQLESLintRule<NamingConventionRuleConfig> = {
           node,
           message: result.errorMessage,
           data: {
+            prefix: prefix,
             suffix: suffix,
             format: style,
             nodeType,
@@ -178,6 +185,7 @@ const rule: GraphQLESLintRule<NamingConventionRuleConfig> = {
       }
       return {
         style: value,
+        prefix: '',
         suffix: '',
       };
     };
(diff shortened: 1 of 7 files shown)


packages/plugin/src/rules/naming-convention.ts
```diff
@@ -8,39 +8,78 @@ const formats = {
...
+  if (ok) {
+    return { ok: true };
+  }
+  return { ok: false, errorMessage: '{{nodeType}} name "{{nodeName}}" should be in {{format}} format' };
 }
 
 const schemaOption = {
-  type: 'string',
-  enum: ['camelCase', 'PascalCase', 'snake_case', 'UPPER_CASE'],
+  type: ['string', 'object'],
```
eddeee888, line 51: Since this is now an array, the enum option [here](https://github.com/dotansimha/graphql-eslint/pull/201/files#diff-9afe7600c3e6002f30eb08d02a06c30afda813baa781871f4b9c7af93039f2d6L24) no longer works and we have to validate  in `checkNameFormat `

dotansimha:
Thanks @eddeee888 ! Sorry for the late reply. I will look into that and review in a few days :)

packages/plugin/src/rules/naming-convention.ts
```diff
@@ -8,39 +8,78 @@ const formats = {
   UPPER_CASE: /^[A-Z_]*$/g,
 };
 
-function checkNameFormat(value, style, leadingUnderscore, trailingUnderscore) {
+type ValidNaming = 'camelCase' | 'PascalCase' | 'snake_case' | 'UPPER_CASE';
+const acceptedStyles: ValidNaming[] = ['camelCase', 'PascalCase', 'snake_case', 'UPPER_CASE'];
```
dotansimha, line 12: You can define the `const` object as array of strings, and then create the `type ValidNaming` like that: `type ValidNaming = typeof acceptedStyles[number]`

packages/plugin/src/rules/naming-convention.ts
```diff
@@ -8,39 +8,78 @@ const formats = {
...
+interface CheckNameFormatParams {
+  value: string;
+  style?: ValidNaming;
+  leadingUnderscore: 'allow' | 'forbid';
+  trailingUnderscore: 'allow' | 'forbid';
+  suffix: string;
```
dotansimha, line 18: Any chance you can add here a `prefix` as well? :)

packages/plugin/src/rules/naming-convention.ts
dotansimha: All good. I'm fine with breaking change at this point.

dotansimha (changes requested):
Looks great, thank you @eddeee888 . A few notes:
1. Please make sure to add a `changeset` (run `yarn changeset` and it will guide you). Please bump `minor`. 
2. We have self-documenting rules now, so I had to make some changes to the rules definition, can you please rebase? (to run the docs generation, use `yarn generate:docs`. 
3.

packages/plugin/src/rules/naming-convention.ts
eddeee888: This is implemented in https://github.com/dotansimha/graphql-eslint/pull/201/commits/8bc38230826356a0efda5946ea83a0b0e9b07439

packages/plugin/src/rules/naming-convention.ts
eddeee888: I have updated it in https://github.com/dotansimha/graphql-eslint/pull/201/commits/d337729c209478609c30f3a531c9bcefab6323fb. Hopefully that is what you meant :)

eddeee888:
Hi @dotansimha, thank you for your feedback! I have made some changes as suggested. Please let me know if you still see problems!

The only thing I'm not 100% sure about is the doc generation. I have run `yarn generate:doc` it but it looks like it's not handling the object case very well. Am I supposed to fill in the missing space?

packages/plugin/src/rules/naming-convention.ts
eddeee888: I'm also thinking that there could be `bannedPrefix` and `bannedSuffix` options as well. Maybe that can go in a separate PR?

dotansimha:
> Hi @dotansimha, thank you for your feedback! I have made some changes as suggested. Please let me know if you still see problems!
> 
> The only thing I'm not 100% sure about is the doc generation. I have run `yarn generate:doc` it but it looks like it's not handling the object case very well. Am I supposed to fill in the missing space?

Thank you so much!

It should be generated correctly, so maybe it's just something with the generation script. Can you please point me to the missing/incorrect generate doc? I will try to fix that.

packages/plugin/src/rules/naming-convention.ts
dotansimha: Yeah, we can add that later, thank you.

packages/plugin/src/rules/naming-convention.ts
dotansimha: Yeah, perfect :) Just to have a single place where the consts are defined.

eddeee888:
Looking at the generated doc file, there are only headers with no expected values / comment in between:
https://github.com/dotansimha/graphql-eslint/blob/1764bca129cef8ac62ab38f9348757b601f5a822/docs/rules/naming-convention.md

Maybe it's how the doc generator works but maybe something like this communicates the config values a bit better :)
```
#### `InputValueDefinition` (string,object)
- style: string, enum
  - `option1`
  - `option2`
- prefix: string
- suffix: string

```

dotansimha:
Oh it should follow JSON schema structure, and I think declaring `type: ['string', 'object'],` along with properties isn't valid in this case. Also, there is a bug in the json-schema-to-markdown we are using (see https://github.com/saibotsivad/json-schema-to-markdown/issues/15). 

It seems like using `definitions` should fix that, but then it doesn't support `enum` correctly. I think eventually we'll have to switch to a different library, but I couldn't find one that is stable/maintained/output single-file. 

I applied a workaround and it seems good enough for now. Thank you @eddeee888 !

dotansimha (approved):

PR #204, opened 2020-12-07
Update avoid-operation-name-prefix.md

aleccool213:


dotansimha:
Awesome catch, thanks @aleccool213 !

PR #231, opened 2020-12-20
Add code file loader when using graphql-config

EmrysMyrddin:
# Purpose

This PR aims to unify the loaders available when using eslint config and `graphql-config`.

Related #232

## Context

Today, `graphql-eslint` is loading schema and operations via `graphql-config` if available and not disable. Otherwise, it uses directly `graphql-tools` and `parserOptions` of eslint configuration file.

## Problem

When parsing schema and operations directly via `graphql-tools`, we use some loaders such as `CodeFileLoader` which allow to parse graphql inside JavaScript and TypeScript files.

But when we use `graphql-config`, this loaders are not used by default.

This lead to projects not lintable if used with graphql-config.

## Solution

My proposition is to create a custom extension for `graphql-config` which allows us to add the same loaders than when we directly use `graphql-tools`.

We don't have to mind if some loaders are already setup by default, `graphql-config` has a check to not register the same loader multiple times.

## What have been changed

I have made a little refactoring: I put loaders lists for schemas and documents in functions to be reusable. I have then used this functions in place of the loaders lists in `loadSiblings` and `loadSchemas`.

I have then created a custom `graphql-config` extension which also use the loaders list and register every loaders in `graphql-config` loaders registry. I didn't found a better way to add custom loaders to `graphql-config` and after some search it seems to be the recommended way.

### DIFF
--- a/packages/plugin/src/parser.ts
+++ b/packages/plugin/src/parser.ts
@@ -13,8 +13,8 @@ export function parse(code: string, options?: ParserOptions): Linter.ESLintParse
 }
 
 const addCodeFileLoaderExtension: GraphQLExtensionDeclaration = (api) => {
-  schemaLoaders().forEach(loader => api.loaders.schema.register(loader))
-  operationsLoaders().forEach(loader => api.loaders.documents.register(loader))
+  schemaLoaders.forEach(loader => api.loaders.schema.register(loader))
+  operationsLoaders.forEach(loader => api.loaders.documents.register(loader))
   return { name: 'graphql-eslint-loaders' }
 }
 
--- a/packages/plugin/src/parser.ts
+++ b/packages/plugin/src/parser.ts
@@ -4,12 +4,18 @@ import { GraphQLError, TypeInfo } from 'graphql';
 import { Linter } from 'eslint';
 import { GraphQLESLintParseResult, ParserOptions } from './types';
 import { extractTokens } from './utils';
-import { getSchema } from './schema';
-import { getSiblingOperations } from './sibling-operations';
-import { loadConfigSync, GraphQLConfig } from 'graphql-config';
+import { getSchema, schemaLoaders } from './schema';
+import { getSiblingOperations, operationsLoaders } from './sibling-operations';
+import { loadConfigSync, GraphQLConfig, GraphQLExtensionDeclaration } from 'graphql-config';
 
 export function parse(code: string, options?: ParserOptions): Linter.ESLintParseResult['ast'] {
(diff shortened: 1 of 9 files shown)


packages/plugin/src/sibling-operations.ts
```diff
@@ -18,6 +18,20 @@ import { dirname, join } from 'path';
 export type FragmentSource = { filePath: string; document: FragmentDefinitionNode };
 export type OperationSource = { filePath: string; document: OperationDefinitionNode };
 
+export const operationsLoaders = (): Loader<string, SingleFileOptions>[] => [
```
dotansimha, line 21: I'm not sure this must be a function that re-creates the instance every time. ESLint is creating the parser services per each file, so it might create tons of instances of each loader.

dotansimha (changes requested):
This is great, thank you @EmrysMyrddin ! Just a small optimizations to make sure we are not creating too many instance of the loaders. 
Also, can you please add a `changeset`? (Run `yarn changeset` and it will guide you).

packages/plugin/src/sibling-operations.ts
EmrysMyrddin: Ok, I don't know if the loaders instances are reusable or should be recreated each time. I will look at the code to see if it can be a problem.

dotansimha: It's reusable, for sure, and it's completely stateless.

dotansimha:
Just a missing `changeset` and we are good to go ;)

packages/plugin/src/parser.ts
```diff
@@ -4,20 +4,27 @@ import { GraphQLError, TypeInfo } from 'graphql';
...
+}
+
 export function parseForESLint(code: string, options?: ParserOptions): GraphQLESLintParseResult {
   const gqlConfig: GraphQLConfig | null = options?.skipGraphQLConfig
     ? null
     : loadConfigSync({
         throwOnEmpty: false,
         throwOnMissing: false,
+        extensions: [addCodeFileLoaderExtension]
       });
```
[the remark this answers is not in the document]
dotansimha: If you are running without graphql-config, we cache those things to avoid performance issues. 
I guess we'll need to do it for graphql-config as well. I created a new issue for it: https://github.com/dotansimha/graphql-eslint/issues/233

dotansimha:
Thank you so much @EmrysMyrddin !

EmrysMyrddin:
Thank you for you responsiveness ! It's a pleasure to work on the repo :-)

Issue #233, opened 2020-12-21
Cache graphql-config object across execution 


dotansimha:
Related: https://github.com/dotansimha/graphql-eslint/pull/231#discussion_r546670947

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@0.7.0`, thanks @EmrysMyrddin !

PR #235, opened 2020-12-21
Add cache for graphql-config loading

EmrysMyrddin:
## Purpose

This PR aims to cache `graphql-config` object to avoid loading configuration for each linted file.

Related #233 

## Context

For now, we are loading the graphql configuration file in the parser, which is called by eslint for each linted file.

## Problem

Since the configuration should not change between the parsing of to files, we shouldn't reload it each time. It is a performance optimization but also a matters of coherence during execution. With a lot of files, there is a timming where we can change the configuration file during eslint parsing phase. Which will lead to have some file linted with old grpahql config, some other with the new one. 

## Solution

My proposition is to globally cache the graphql config object for the entire execution.

## What have been changed

I have moved out graphql-config logic to its own file which allow me to have a private mutable variable to store the cache of the configuration. Say me if you prefer to have this back in `parser.ts`.

I have also refactored some correlated code to make it more clear and remove redundant checks. Same thing, don't hesitate to make suggestions with your are not agree, I think those are more subjective changes.

### DIFF
--- a/packages/plugin/src/schema.ts
+++ b/packages/plugin/src/schema.ts
@@ -23,26 +23,24 @@ export const schemaLoaders: Loader<string, SingleFileOptions>[] = [
   new GraphQLFileLoader(),
   new JsonFileLoader(),
   new UrlLoader(),
-]
+];
 
 export function getSchema(options: ParserOptions, gqlConfig: GraphQLConfig): GraphQLSchema | null {
   let schema: GraphQLSchema | null = null;
 
   // We first try to use graphql-config for loading the schema, based on the type of the file,
   // We are using the directory of the file as the key for the schema caching, to avoid reloading of the schema.
-  if (options && options.filePath && !options.skipGraphQLConfig) {
+  if (gqlConfig && options?.filePath) {
     const fileDir = dirname(options.filePath);
 
     if (schemaCache.has(fileDir)) {
       schema = schemaCache.get(fileDir);
     } else {
-      if (gqlConfig) {
-        const projectForFile = gqlConfig.getProjectForFile(options.filePath);
+      const projectForFile = gqlConfig.getProjectForFile(options.filePath);
 
-        if (projectForFile) {
-          schema = projectForFile.getSchemaSync();
-          schemaCache.set(fileDir, schema);
-        }
+      if (projectForFile) {
+        schema = projectForFile.getSchemaSync();
+        schemaCache.set(fileDir, schema);
       }
     }
   }
@@ -50,7 +48,7 @@ export function getSchema(options: ParserOptions, gqlConfig: GraphQLConfig): Gra
   // If schema was not loaded yet, and user configured it in the parserConfig, we can try to load it,
   // In this case, the cache key is the path for the schema. This is needed in order to allow separate
   // configurations for different file paths (a very edgey case).
-  if (options && !schema && options.schema) {
+  if (!schema && options?.schema) {
     const schemaKey = Array.isArray(options.schema) ? options.schema.join(',') : options.schema;
 
     if (schemaCache.has(schemaKey)) {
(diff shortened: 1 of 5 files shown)


packages/plugin/src/schema.ts
```diff
@@ -30,27 +30,25 @@ export function getSchema(options: ParserOptions, gqlConfig: GraphQLConfig): Gra
 
   // We first try to use graphql-config for loading the schema, based on the type of the file,
   // We are using the directory of the file as the key for the schema caching, to avoid reloading of the schema.
-  if (options && options.filePath && !options.skipGraphQLConfig) {
+  if (gqlConfig && options?.filePath) {
```
EmrysMyrddin, line 33: Here, we don't have to check for `options.skipGraphqlConfig` since `graphqlConfig` is null this case.

packages/plugin/src/schema.ts
```diff
@@ -30,27 +30,25 @@ export function getSchema(options: ParserOptions, gqlConfig: GraphQLConfig): Gra
...
     const fileDir = dirname(options.filePath);
 
     if (schemaCache.has(fileDir)) {
       schema = schemaCache.get(fileDir);
     } else {
-      if (gqlConfig) {
```
EmrysMyrddin, line 39: Here, we can check for this earlier, we can't have something in the cache if we don't even have a graphql config.

packages/plugin/src/sibling-operations.ts
```diff
@@ -60,32 +60,30 @@ export function getSiblingOperations(options: ParserOptions, gqlConfig: GraphQLC
 
   // We first try to use graphql-config for loading the operations paths, based on the type of the file,
   // We are using the directory of the file as the key for the schema caching, to avoid reloading of the schema.
-  if (options && options.filePath && !options.skipGraphQLConfig) {
+  if (gqlConfig && options?.filePath) {
```
EmrysMyrddin, line 63: Same here, by checking for `options.skipGraphQLConfig` and `gqlConfig` is redundant.

packages/plugin/src/schema.ts
dotansimha: You are right, but there is a specific use case, where you install eslint and this plugin in a monorepo that has graphql-config in the root. This will force graphql-eslint to use graphql-config even if you don't want to. I added this for flexibility and for allowing to explicitly skip graphql-config.

packages/plugin/src/sibling-operations.ts
dotansimha: Please see my comment above, we need to have the ability to explicitly skip graphql-config.

packages/plugin/src/schema.ts
EmrysMyrddin: I don't understand :-)

In `loadGraphqlConfig`, there is a check that returns `null` if `options.skipGraphQLConfig` is true.

Then, the loaded `gqlConfig` (`null` in our case) will be passed along with `options` to `getSchema`. So here, I don't understand how we can have both `gqlConfig` not `null` and `options.skipGraphQLConfig` be `true` ?

dotansimha: So there are 2 possible flows here:
1. If you have `graphql-config` and you wish to use it - `loadGraphqlConfig` will load and return the parsed graphql-config object. Then, when it's passes to `getSchema` - it knows to use it, instead of using the `option.schema` configuration. 
2. If you have `graphql-config` and don't want `graphql-eslint` to use it - you set `skipGraphQLConfig` and it will use `null` for the graphql-config object, like it was not found or not available (basically, fallbacks to 3)
3. If you have `null` in `gqlConfig` => we are going to use `@graphql-tools` directly to load the schema and the operations, instead of using graphql-config.

packages/plugin/src/schema.ts
dotansimha: Perfectly right, thanks!

dotansimha:
I think you might need a tiny rebase and a `changeset` and we are good to go ;)

dotansimha (approved):

Issue #237, opened 2020-12-21
Cache Sibblings Operations object


EmrysMyrddin:
For now, sibblings operations source code extract is cached. But the SibblingsOperations object is recreated each time we get the sibblings (for each linted file).

This means that the fragments and operations cache is valid only across on file, not across the entire execution.

Is this intended ?

https://github.com/dotansimha/graphql-eslint/blob/d3ff768511cb6de4da3e5104fe072f2c2ca3923f/packages/plugin/src/sibling-operations.ts#L121

dotansimha:
Thank you for reporting it @EmrysMyrddin ! 

I think maybe the implementation there is a bit confusing. 
There is `operationsCache` (defined here: https://github.com/dotansimha/graphql-eslint/blob/d3ff768511cb6de4da3e5104fe072f2c2ca3923f/packages/plugin/src/sibling-operations.ts#L56 ) which is the cache for the entire siblings operations, where the key is the execution directory, and the values is the result of the lookup.
This piece of code is in charge of making sure that we don't run graphql-config or graphql-tools for searching the actual file-system for operations.
This is done at the level of the parser, and (hopefully) happens once, it keeps a `Source` objects and doesn't do any AST traverse/visit/manipulation yet, so we don't really know what we have inside the found operation files.

Then, when a rule asks for the siblings operations, we have an additional cache (`fragmentsCache` and `cachedOperations`) which intended to make sure that even if a rule calls `getFragments` multiple times, we do the AST visiting for finding the fragments (or, operations) only once. This is just an optimization that is intended to make sure we don't visit the found operations multiple times during the same execution. 

Does it makes sense? because I might miss something here, and this could be optimized.

EmrysMyrddin:
From what I understand, `getSiblingOperations` is called by the parser, which is called once by file linted (so multiple times by execution). We are caching siblings `Source` list in `operationsCache`, which is a global cache across all execution.

But, starting from line 119, we no longer are storing data in global cache but local cache. We recreate all the functions returned by `getSiblingOperations` and the cache variables `fragmentsCache` and `cachedOperations`. This means that the AST visit will be possibly done once by file linted, not once by execution.

I'm not sure I'm very clear 😛 I'm not a native english speaker :-)

EmrysMyrddin:
I have opened a draft PR to show you what I'm talking about: https://github.com/dotansimha/graphql-eslint/pull/240 

Here I have added an execution level cache for AST visit. Is it more clear ?

[...]

PR #240, opened 2020-12-21
:alembic: explanation for issue #237

EmrysMyrddin:


dotansimha:
It makes perfect sense, you are right, thank you @EmrysMyrddin :) 
I think this can be converted into a PR to change that?

Issue #155, opened 2020-11-17
New rule: no-unreachable


kamilkisiela:
```graphql
type Query {
  foo: String
}

type Unreachable {
   id: ID
}
```

dotansimha:
Available in `@graphql-eslint/eslint-plugin@0.7.0`. Thanks @kamilkisiela

PR #243, opened 2020-12-22
NEW RULE: no-unreachable-types rule

kamilkisiela:
Related #155

### DIFF
--- a/packages/plugin/src/parser.ts
+++ b/packages/plugin/src/parser.ts
@@ -7,7 +7,7 @@ import { extractTokens } from './utils';
 import { getSchema } from './schema';
 import { getSiblingOperations } from './sibling-operations';
 import { loadGraphqlConfig } from './graphql-config';
-import { getReachableTypes } from './graphql-ast';
+import { createReachableTypesService } from './graphql-ast';
 
 export function parse(code: string, options?: ParserOptions): Linter.ESLintParseResult['ast'] {
   return parseForESLint(code, options).ast;
@@ -16,13 +16,12 @@ export function parse(code: string, options?: ParserOptions): Linter.ESLintParse
 export function parseForESLint(code: string, options?: ParserOptions): GraphQLESLintParseResult {
   const gqlConfig = loadGraphqlConfig(options);
   const schema = getSchema(options, gqlConfig);
-  const reachableTypes = getReachableTypes(schema);
   const siblingOperations = getSiblingOperations(options, gqlConfig);
   const parserServices = {
     hasTypeInfo: schema !== null,
     schema,
     siblingOperations,
-    reachableTypes,
+    getReachableTypes: createReachableTypesService(schema),
   };
 
   try {
--- a/packages/plugin/src/parser.ts
+++ b/packages/plugin/src/parser.ts
@@ -7,6 +7,7 @@ import { extractTokens } from './utils';
 import { getSchema } from './schema';
 import { getSiblingOperations } from './sibling-operations';
 import { loadGraphqlConfig } from './graphql-config';
+import { getReachableTypes } from './graphql-ast';
 
 export function parse(code: string, options?: ParserOptions): Linter.ESLintParseResult['ast'] {
   return parseForESLint(code, options).ast;
@@ -15,11 +16,13 @@ export function parse(code: string, options?: ParserOptions): Linter.ESLintParse
 export function parseForESLint(code: string, options?: ParserOptions): GraphQLESLintParseResult {
   const gqlConfig = loadGraphqlConfig(options);
   const schema = getSchema(options, gqlConfig);
+  const reachableTypes = getReachableTypes(schema);
   const siblingOperations = getSiblingOperations(options, gqlConfig);
   const parserServices = {
     hasTypeInfo: schema !== null,
     schema,
     siblingOperations,
+    reachableTypes,
   };
 
   try {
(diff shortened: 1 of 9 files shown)


packages/plugin/src/rules/no-unreachable-types.ts
```diff
@@ -0,0 +1,80 @@
+import { GraphQLESLintRule } from '../types';
+import { requireReachableTypesFromContext } from '../utils';
+
+const UNREACHABLE_TYPE = 'UNREACHABLE_TYPE';
+
+const rule: GraphQLESLintRule = {
+  meta: {
+    messages: {
+      [UNREACHABLE_TYPE]: `Type "{{ typeName }}" in unreachable`,
```
dotansimha, line 9: is* 
Do we need different message for other graphql kinds? like `{{ kind }} {{ name }} is unreachable` ? so it will be `scalar Email is unreachable by the loaded operations`.

dotansimha (changes requested):

packages/plugin/src/parser.ts
```diff
@@ -15,11 +16,13 @@ export function parse(code: string, options?: ParserOptions): Linter.ESLintParse
 export function parseForESLint(code: string, options?: ParserOptions): GraphQLESLintParseResult {
   const gqlConfig = loadGraphqlConfig(options);
   const schema = getSchema(options, gqlConfig);
+  const reachableTypes = getReachableTypes(schema);
```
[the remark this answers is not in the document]
kamilkisiela: but we've got no caching for such things

packages/plugin/src/parser.ts
```diff
@@ -15,11 +16,13 @@ export function parse(code: string, options?: ParserOptions): Linter.ESLintParse
...
   const siblingOperations = getSiblingOperations(options, gqlConfig);
   const parserServices = {
     hasTypeInfo: schema !== null,
```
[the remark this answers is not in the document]
kamilkisiela: I didn't touch `hasTypeInfo`

packages/plugin/src/rules/no-unreachable-types.ts
kamilkisiela, line 9:
```suggestion
      [UNREACHABLE_TYPE]: `Type "{{ typeName }}" is unreachable`,
```

packages/plugin/tests/no-unreachable-types.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        type User {
+          id: ID
+          name: String
+        }
+      `),
+      errors: [
+        {
+          message: `Type "User" in unreachable`,
```
kamilkisiela, line 48:
```suggestion
          message: `Type "User" is unreachable`,
```

packages/plugin/tests/no-unreachable-types.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+          name: String
+        }
+
+        input UsersFilter {
+          limit: Int
+        }
+      `),
+      errors: [
+        {
+          message: `Type "UsersFilter" in unreachable`,
```
kamilkisiela, line 69:
```suggestion
          message: `Type "UsersFilter" is unreachable`,
```

packages/plugin/tests/no-unreachable-types.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+
+        type Articles {
+          id: ID
+          title: String
+          author: User
+        }
+      `),
+      errors: [
+        {
+          message: `Type "Articles" in unreachable`,
```
kamilkisiela, line 92:
```suggestion
          message: `Type "Articles" is unreachable`,
```

packages/plugin/tests/no-unreachable-types.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        type Comment {
+          id: ID
+          article: Article
+        }
+      `),
+      errors: [
+        {
+          message: `Type "Comment" in unreachable`,
```
kamilkisiela, line 121:
```suggestion
          message: `Type "Comment" is unreachable`,
```

packages/plugin/src/rules/no-unreachable-types.ts
kamilkisiela: This rule is not about types being unused by operations, it's all within schema. Everything in schema starts with in fields of Query, Mutation or Subscription. The type is reachable if you can access it in those fields directly or indirectly any level deep. Take a look at the tests.

kamilkisiela: There's another rule we can implement, something that I do in GraphQL Inspector which involves operations and fragments to detect unused pieces of schema.

packages/plugin/src/parser.ts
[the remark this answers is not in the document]
kamilkisiela: Now it's lazy and cached

packages/plugin/src/rules/no-unreachable-types.ts
kamilkisiela: @dotansimha Scalar is a type so I think it's fine to use just `type`. We don't need more messages because we're tracking only type definitions and extensions, not their fields and values.

kamilkisiela:
@dotansimha added auto-fixer

Issue #253, opened 2021-01-03
[naming-convention] Add Query-specific option


eddeee888:
I think there could be a new option that targets the queries because the naming convention could be different from a field definition on a normal type. 

For example, I may want to forbid the queries to have `get` prefixes because queries are always "GET" but still allows other object type definitions, then I can add the following rule to eslint config:

```json
"@graphql-eslint/naming-convention": [
  "error",
  {
    "QueryFieldDefinition": { "format": "camelCase", "forbiddenPrefixes": ["get"] },
    "FieldDefinition": { "format": "camelCase" },
  }
]
```

```graphql
type Query {
  getUser: User # Error
  user: User # OK
}

type User {
  getter: User # OK
}
```

If it makes sense I can try creating a PR for this :)

dotansimha:
We have this: https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/avoid-operation-name-prefix.md :) 

```
@graphql-eslint/avoid-operation-name-prefix: ["error", [{"keywords":["get"]}]]
```

eddeee888:
I had a look and I think [avoid-operation-name-prefix](https://github.com/dotansimha/graphql-eslint/blob/d9e93779ea76fa112af0f5d2fb21324b155a209d/packages/plugin/src/rules/avoid-operation-name-prefix.ts#L103-L108) only works for `OperationDefinition` and `FragmentDefinition` at the moment? 
I think it would be nice to be able to enforce a similar rule to query definitions in the schema files as well :)

dotansimha:
Oh, that totally makes sense, mixed the two :) Re-opening :)

eddeee888:
Awesome thanks! I should have made it clearer that it's for schema files. I'll look into this and create a PR soon :)

eddeee888:
@dotansimha I have created a PR for this: https://github.com/dotansimha/graphql-eslint/pull/256
Please have a look when you have a chance, thanks! 🙏

dotansimha:
Available in `@graphql-eslint/eslint-plugin@0.8.0`

PR #256, opened 2021-01-05
[naming-convention]  Add QueryDefinition option, forbiddenSuffixes and forbiddenPrefixes properties.

eddeee888:
This PR implements the following:

- `QueryDefinition` option: This can be used specifically to target queries. As a result, the `FieldDefinition` option no longer targets queries which could be a breaking change
- `forbiddenPrefixes` property: used to report if any definition has one of the declared prefixes
- `forbiddenSuffixes` property: used to report if any definition has one of the declared suffixes

Other minor change(s):

- `checkNode` now takes full `property` object to make future properties easier to implement
- Fix wrong node type for UnionDefinition [here](https://github.com/dotansimha/graphql-eslint/pull/256/files#diff-9afe7600c3e6002f30eb08d02a06c30afda813baa781871f4b9c7af93039f2d6R345). Was `Scalar`, now `Union`

Related: https://github.com/dotansimha/graphql-eslint/issues/253

### DIFF
--- a/packages/plugin/src/rules/naming-convention.ts
+++ b/packages/plugin/src/rules/naming-convention.ts
@@ -23,9 +23,19 @@ interface CheckNameFormatParams {
   prefix: string;
   suffix: string;
   forbiddenPrefixes: string[];
+  forbiddenSuffixes: string[];
 }
 function checkNameFormat(params: CheckNameFormatParams): { ok: false; errorMessage: string } | { ok: true } {
-  const { value, style, leadingUnderscore, trailingUnderscore, suffix, prefix, forbiddenPrefixes } = params;
+  const {
+    value,
+    style,
+    leadingUnderscore,
+    trailingUnderscore,
+    suffix,
+    prefix,
+    forbiddenPrefixes,
+    forbiddenSuffixes,
+  } = params;
   let name = value;
   if (leadingUnderscore === 'allow') {
     [, name] = name.match(/^_*(.*)$/);
@@ -55,6 +65,14 @@ function checkNameFormat(params: CheckNameFormatParams): { ok: false; errorMessa
     };
   }
 
+  if (forbiddenSuffixes.some(forbiddenSuffix => name.endsWith(forbiddenSuffix))) {
+    return {
+      ok: false,
+      errorMessage:
+        '{{nodeType}} "{{nodeName}}" should not have one of the following suffix(es): {{forbiddenSuffixes}}',
+    };
+  }
+
   if (!formats[style]) {
     return { ok: true };
   }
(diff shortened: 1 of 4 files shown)


packages/plugin/src/rules/naming-convention.ts
```diff
@@ -225,61 +283,67 @@ const rule: GraphQLESLintRule<NamingConventionRuleConfig> = {
...
+        // TODO: check proper node typing
+        if (options.QueryDefinition && isQueryType(node.parent)) {
+          const property = normalisePropertyOption(options.QueryDefinition);
+          checkNode(node.name, property, 'Query');
+        }
+
+        if (options.FieldDefinition && !isQueryType(node.parent)) {
           const property = normalisePropertyOption(options.FieldDefinition);
-          checkNode(node.name, property.style, 'Field', property.suffix, property.prefix);
+          checkNode(node.name, property, 'Field');
```
eddeee888, line 316: As mentioned in the description, this could be a breaking change because `FieldDefinition` no longer targets queries

packages/plugin/src/rules/naming-convention.ts
```diff
@@ -225,61 +283,67 @@ const rule: GraphQLESLintRule<NamingConventionRuleConfig> = {
...
       InputObjectTypeDefinition: node => {
         if (options.InputObjectTypeDefinition) {
           const property = normalisePropertyOption(options.InputObjectTypeDefinition);
-          checkNode(node.name, property.style, 'Input type', property.suffix, property.prefix);
+          checkNode(node.name, property, 'Input type');
         }
       },
-      FieldDefinition: node => {
-        if (options.FieldDefinition) {
+      FieldDefinition: (node: any) => {
```
eddeee888, line 307: I think the `node` type in  [typekit](https://github.com/dotansimha/graphql-eslint/blob/9899c0b1e824cdda9835b22893459853e1242cf9/packages/plugin/src/testkit.ts#L9) could be updated to contain `parent`:

```
export type GraphQLESlintRuleListener<WithTypeInfo extends boolean> = {
  [K in keyof ASTKindToNode]?: (
    node: GraphQLESTreeNode<ASTKindToNode[K], WithTypeInfo> & {
      parent: GraphQLESTreeNode<ASTKindToNode[K], WithTypeInfo>;
    }
  ) => void;
} &
  Record<string, any>;
```

However, I don't know what effect it has in other places so I'll let you decide whether I should update it :)

eddeee888:
Hi @dotansimha , can you please take a look when you have a chance :) ?

packages/plugin/src/rules/naming-convention.ts
dotansimha: That's fine :) I think it's a fair change. We accept breaking changes until v1, so it's not a big deal at the moment.

dotansimha:
Looks great. Thank you @eddeee888 !

Issue #267, opened 2021-01-17
bug: Definition for rule 'no-unreachable-types' was not found 


dnalborczyk:
version: **0.7.1**

I believe the file is not being exported: https://github.com/dotansimha/graphql-eslint/blob/%40graphql-eslint/eslint-plugin%400.7.1/packages/plugin/src/rules/index.ts

a doc reference seems to be missing in the rules overview as well.

I'll try to put up a PR: https://github.com/dotansimha/graphql-eslint/pull/268

dotansimha:
Thank you @dnalborczyk , will release a new version in a few minutes :)

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@0.7.2` :)

dnalborczyk:
thank you @dotansimha !!

PR #268, opened 2021-01-17
fix: Definition for rule 'no-unreachable-types' was not found

dnalborczyk:
related: https://github.com/dotansimha/graphql-eslint/issues/267


dotansimha:
Thank you @dnalborczyk !

Issue #276, opened 2021-01-21
input-name rule option to specify casing


lioneltay:
At the moment, when using the `input-name` rule https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/input-name.md, this is an error

```graphql
# eslint @graphql-eslint/input-name: ["error", [{"checkInputType":true}]]

type Mutation {
  setMessage(input: SetMessageInput): String
}
```

While this passes

```graphql
# eslint @graphql-eslint/input-name: ["error", [{"checkInputType":true}]]

type Mutation {
  SetMessage(input: SetMessageInput): String
}
```

Could an option be added to allow the mutation name to still be in camel case, while the type name should be the PascalCase version appended with `Input`?

Maybe this will allow this rule to work better with the `naming-convention` rule https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/naming-convention.md.

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@0.8.0`

PR #286, opened 2021-01-28
Add options to input-name rule. Addresses #276

lioneltay:
Add caseSensitiveInputType, checkMutations, checkQueries options to input-name rule

caseSensitiveInputType: When false the input type name must be <mutationName>Input but is not case sensitive.

checkMutations (default true): Apply this rule to mutations.

checkQueries (default false): Apply this rule to queries.

Setup default option values, previously all defaulting to undefined.

Also fixed documentation error which says that `checkInputType` defaults to `true`, when infact it defaults to `false`.

Addressed https://github.com/dotansimha/graphql-eslint/issues/276

### DIFF
--- a/.changeset/early-turkeys-admire.md
+++ b/.changeset/early-turkeys-admire.md
@@ -1,0 +1,5 @@
+---
+'@graphql-eslint/eslint-plugin': minor
+---
+
+Add options to input-name rule.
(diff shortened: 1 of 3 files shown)


dotansimha:
Amazing work. thank you @lioneltay !

Issue #317, opened 2021-02-16
no-hashtag-description does not allow for eslint directives


ilyavolodin:
I can't seem to get `no-hashtag-description` rule to allow any hashtag comments at all, including eslint directives. Here are two examples that are erroring out for me:
```graphql
# eslint-disable @graphql-eslint/no-unreachable-types

"Types of 'unlessContainsTypes' omit"
enum OmitTypes {
  "Scalar fields"
  scalar
  "Complex type fields"
  nonScalar
}
```

```graphql
# eslint-disable-next-line @graphql-eslint/description-style
"""
The different roles a funding instrument can have, depending on different features that Venmo offers.
Usually it's either used as the default instrument, as a backup if the default cannot be used, or in some features,
it indicates if the funding instrument can be used for that feature
"""
type FundingInstrumentRoles {
...
}
```

I would suggest to modify the rule to exclude any hashtag comments that contain eslint directives, i.e. `# eslint` should always be excluded from this rule.

dotansimha:
Sure, totally makes sense :) Thanks @ilyavolodin

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@0.8.1`

PR #318, opened 2021-02-17
no-hashtag-description allow eslint directives

gilgardosh:
related #317


packages/plugin/src/utils.ts
```diff
@@ -99,3 +99,14 @@ export function extractTokens(source: string): AST.Token[] {
 
   return tokens;
 }
+
+export function checkForEslint(token: Token, rawNode: DocumentNode): boolean {
+  if (token.kind !== 'Comment') {
+    return false;
+  }
+  const string = rawNode.loc.source.body.substring(token.start+1, token.start+8);
```
dotansimha, line 107: @gilgardosh the `loc` might be optional. Can you please add optional chaining here?

Issue #315, opened 2021-02-16
`@graphql-eslint/no-unreachable-types` reports issues with interfaces


dotansimha:
Given this

```
interface User {

}

type SuperUser implements User {

}

type Query {
  user: User!
}
```

We'll get an error regarding `SuperUser` no in use.

dnalborczyk:
@dotansimha seems to be correct. `SuperUser` is not connected to the root. I thought that would be the whole point of this linting rule?

dotansimha:
@dnalborczyk but it does implementing the `User`. And GraphQL schema can declare the interface as return type, and the actual type is defined during resolvers execution. @kamilkisiela what do you think?

kamilkisiela:
Yeah, I need to fix it.

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@0.8.1`

ostrowr:
I can open another issue if helpful, but I think this one should be reopened. The linked PR solves the issue for types that are only reachable directly from an interface, but not for one that is available in an implementation of the interface.

Using the example above:

```
interface User {
    id: String
}

type SuperUser implements User {
    id: String 
    superDetail: SuperDetail
}

# This one is still marked as unreachable!
type SuperDetail {
    detail: String
}

type Query {
  user: User!
}
```

Additionally, 

`extend type SuperUser...`

is valid GraphQL but will also, strangely, be marked as unreachable even though SuperUser is marked as reachable.

kamilkisiela:
I'll fix it

kamilkisiela:
#361

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@0.9.1`. Thanks @kamilkisiela !

PR #319, opened 2021-02-17
Pass if type implements reachable interface

gilgardosh:
related: #315


kamilkisiela (approved):
I added few more tests

PR #329, opened 2021-02-21
Update naming-convention.md

bradleyayers:
Remove extraneous array wrapping.

### DIFF
--- a/docs/rules/naming-convention.md
+++ b/docs/rules/naming-convention.md
@@ -12,7 +12,7 @@ Requires description around GraphQL nodes
 ### Incorrect
 
 ```graphql
-# eslint @graphql-eslint/naming-convention: ["error", [{"ObjectTypeDefinition":"PascalCase"}]]
+# eslint @graphql-eslint/naming-convention: ["error", {"ObjectTypeDefinition":"PascalCase"}]
 
 type someTypeName {
   f: String!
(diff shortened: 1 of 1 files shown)


dotansimha:
Hi @bradleyayers ! 
Our documentation is generated automatically based on the JSON schema defined in each rule, so the documentation schema for this specific rule is coming from here: https://github.com/dotansimha/graphql-eslint/blob/master/packages/plugin/src/rules/naming-convention.ts#L148 

I assume this was opened because you had an issue with the generated example? Can you please elaborate on this? 

Thanks!

bradleyayers:
Hey @dotansimha, so the problem is that the example is wrong (it's invalid if you try to actually use it):

Instead of

```json
["error", [{"ObjectTypeDefinition":"PascalCase"}]]
```

it should be

```json
["error", {"ObjectTypeDefinition":"PascalCase"}]
```

So it sounds like there's an issue with the generator.

B2o5T:
@dotansimha I fixed JSON schema in this PR https://github.com/dotansimha/graphql-eslint/pull/411 so you can close this one because @bradleyayers fixed markdown code and not source code

dotansimha:
Fixed in https://github.com/dotansimha/graphql-eslint/pull/411

Issue #326, opened 2021-02-21
New rule: Strict id in types in schema


eddeee888:
**Is your feature request related to a problem? Please describe.**

This is a feature request for a new rule that targets schema files.

I think when it comes to GraphQL schema design, things that are domain models e.g. `User`, `Post`, `Comment` should always have a field with name `id` and type `ID` in the schema ( these are normally the `id` in the database ). This will help caching according to the [GraphQL spec](https://spec.graphql.org/June2018/#sec-ID)

For result payload types or types that do not have a natural id field such as pagination, it is fine not to have an id.

**Describe the solution you'd like**

It would be good to have a new rule `strict-id-types-in-schema` ( open for name suggestions )

This rule should:

- enforce types to have a field with an `ID` type
- look for an `id` field with `ID` type by default. The name of the field could be changed to other values e.g. `key`
- have an ability to exclude object types that have certain suffix/prefix e.g. `Result`, `Payload`, `Pagination`

**Describe alternatives you've considered**

I've looked at all the existing rules but can't find any that specifically targets `id` fields on the schema

**Additional context**

Happy to put together a PR if this sounds like a good rule :)

dotansimha:
Thank you for reporting this @eddeee888 . This sounds like a fairly custom behavior, but we do follow this kind of structure in our GraphQL projects, so it might make sense to others as well. 
I guess as long as it's easily customizable - this could be a great rule :) PRs are always welcome!

eddeee888:
Hi @dotansimha , this is implemented in https://github.com/dotansimha/graphql-eslint/pull/333. Please have a look when you have a chance 🙏

dotansimha:
Awesome work, @eddeee888 . Thank you! 
It's availble now in `@graphql-eslint/eslint-plugin@0.9.0`.

PR #333, opened 2021-02-27
[New rule] Add strict-id-in-types rule

eddeee888:
## Description

This PR adds a new rule `strict-id-in-types`. 

Usually, a GraphQL output type can be mapped to a domain model e.g. `User`, `Post`, `Comment`. These fields should have a logical database `id` which could be used as the unique identifier of the correspondent GraphQL output type. There are some exceptions: Payload/Result, Pagination, or types used to group related fields together.

This rule enforces all output types to have a unique identifier field unless belonging to the exceptions. This ensures the consistency of output types in the schema.

The default name is `id` and the default type is `ID`. A lot of GraphQL clients e.g. Apollo client and urql uses the `id` field of incoming data to normalise the cache so consistency from the schema level will help writing documents as well.

### Options

- `acceptedIdNames`: an array of strings containing field names that could be used as unique identifier. Works in tandem with `acceptedIdTypes`. Default: `["id"]`
- `acceptedIdTypes`: an array of strings containing field types that could be used as unique identifier. Works in tandem with `acceptedIdNames`. Default: `["ID"]`
- `exceptions`: an object containing properties of a type to be ignored.
  - `suffixes`: Any type that has one of the options declared here should be ignored. Default: `[]`

Related https://github.com/dotansimha/graphql-eslint/issues/326

## Type of change

N/A

## How Has This Been Tested?

I have added the tests in this file: https://github.com/eddeee888/graphql-eslint/blob/2f7efe34967cf26711897bcc614461eb3aa788c0/packages/plugin/tests/strict-id-in-types.spec.ts
To test, run `yarn test strict-id-in-types`

**Test Environment**:
- OS: MacOS
- `@graphql-eslint/...`: Latest versions
- NodeJS: 12.13.1

## Checklist:

## Further comments

- When running `yarn generate:docs`, it changed a lot of unrelated doc files. This makes me think that the latest version on `master` may not have run generate:docs? 🤔  Please let me know how we should deal with this issue.
- It would be good to try this on an alpha version for some of my projects before releasing. Please let me know if it's possible :)

### DIFF
--- a/docs/rules/provided-required-arguments.md
+++ b/docs/rules/provided-required-arguments.md
@@ -7,4 +7,4 @@
 
 A field or directive is only valid if all required (non-null without a default value) field arguments have been provided.
 
-> This rule is a wrapper around a `graphql-js` validation function. [You can find it's source code here](https://github.com/graphql/graphql-js/blob/master/src/validation/rules/ProvidedRequiredArgumentsRule.js).
+> This rule is a wrapper around a `graphql-js` validation function. [You can find it's source code here](https://github.com/graphql/graphql-js/blob/master/src/validation/rules/ProvidedRequiredArguments.js).
--- a/packages/plugin/src/rules/strict-id-in-types.ts
+++ b/packages/plugin/src/rules/strict-id-in-types.ts
@@ -30,8 +30,60 @@ const shouldIgnoreNode = ({ node, exceptions }: ShouldIgnoreNodeParams): boolean
 
 const rule: GraphQLESLintRule<StrictIdInTypesRuleConfig> = {
   meta: {
+    type: 'suggestion',
     docs: {
+      description:
+        'Requires output types to have one unique identifier unless they do not have a logical one. Exceptions can be used to ignore output types that do not have unique identifiers.',
+      category: 'Best practices',
+      recommended: true,
+      requiresSchema: false,
+      requiresSiblings: false,
       url: 'https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/strict-id-in-types.md',
+      examples: [
+        {
+          title: 'Incorrect',
+          usage: [{ acceptedIdNames: ['id', '_id'], acceptedIdTypes: ['ID'], exceptions: { suffixes: ['Payload'] } }],
+          code: /* GraphQL */ `
+            # Incorrect field name
+            type InvalidFieldName {
+              key: ID!
+            }
+
+            # Incorrect field type
+            type InvalidFieldType {
+              id: String!
+            }
+
+            # Incorrect exception suffix
+            type InvalidSuffixResult {
+              data: String!
+            }
+
+            # Too many unique identifiers. Must only contain one.
+            type InvalidFieldName {
+              id: ID!
+              _id: ID!
+            }
+          `,
+        },
+        {
+          title: 'Correct',
+          usage: [{ acceptedIdNames: ['id', '_id'], acceptedIdTypes: ['ID'], exceptions: { suffixes: ['Payload'] } }],
+          code: /* GraphQL */ `
+            type User {
+              id: ID!
+            }
+
+            type Post {
+              _id: ID!
+            }
+
+            type CreateUserPayload {
+              data: String!
+            }
+          `,
+        },
+      ],
     },
   },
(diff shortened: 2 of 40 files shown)


docs/rules/provided-required-arguments.md
```diff
@@ -7,4 +7,4 @@
 
 A field or directive is only valid if all required (non-null without a default value) field arguments have been provided.
 
-> This rule is a wrapper around a `graphql-js` validation function. [You can find it's source code here](https://github.com/graphql/graphql-js/blob/master/src/validation/rules/ProvidedRequiredArgumentsRule.js).
\ No newline at end of file
+> This rule is a wrapper around a `graphql-js` validation function. [You can find it's source code here](https://github.com/graphql/graphql-js/blob/master/src/validation/rules/ProvidedRequiredArguments.js).
```
eddeee888, line 10: A lot of these changes are unrelated to my change. Probably the version on `master` has not generated docs? 🤔

packages/plugin/src/rules/strict-id-in-types.ts
```diff
@@ -0,0 +1,172 @@
...
+        const validIds = node.fields.filter(field => {
+          const fieldNode = field.rawNode();
+
+          const isValidIdName = options.acceptedIdNames.includes(fieldNode.name.value);
+
+          // To be a valid type, it must be non-null and one of the accepted types.
+          let isValidIdType = false;
+          if (fieldNode.type.kind === 'NonNullType' && fieldNode.type.type.kind === 'NamedType') {
+            isValidIdType = options.acceptedIdTypes.includes(fieldNode.type.type.name.value);
+          }
```
eddeee888, line 147: Is there a better way to check node types using some kind of helper from the `graphql` package? I tried so many approaches but couldn't get this working so I went with this. :)

I'm hoping to get some advice/guidance here 🙏

packages/plugin/src/rules/strict-id-in-types.ts
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        const validIds = node.fields.filter(field => {
+          const fieldNode = field.rawNode();
+
+          const isValidIdName = options.acceptedIdNames.includes(fieldNode.name.value);
+
+          // To be a valid type, it must be non-null and one of the accepted types.
+          let isValidIdType = false;
+          if (fieldNode.type.kind === 'NonNullType' && fieldNode.type.type.kind === 'NamedType') {
```
eddeee888, line 145: Is there a better way to check node types using some kind of helper from the `graphql` package? I tried so many approaches but couldn't get this working so I went with this. :)

I'm hoping to get some advice/guidance here 🙏

docs/rules/provided-required-arguments.md
dotansimha: Yeah, maybe it's not fully updated. just ignore it for now :)

packages/plugin/src/rules/strict-id-in-types.ts
dotansimha: `graphql.js` has `isNonNullType` but it's not for AST, it's only for class types. 

I'm totally fine with the way you tested it, but maybe you can use `Kind.NON_NULL_TYPE` and `Kind.NAMED_TYPE` instead of the explicit string valus?

dotansimha (approved):
Looks great, thank you @eddeee888 ! Please see my minor comment, other than that, it looks perfect!

dotansimha:
Oh and one more thing: Maybe we can add additional fields for `exceptions`, something like `types:` to specify exclusions with the complete type name?

packages/plugin/src/rules/strict-id-in-types.ts
eddeee888: Makes sense! This is done in https://github.com/dotansimha/graphql-eslint/pull/333/commits/6ba418a98de2823f98a605e117f034fb12090d39. Cheers!

eddeee888:
Good idea @dotansimha ! 👍 

`exceptions.types` is implemented in https://github.com/dotansimha/graphql-eslint/pull/333/commits/4c79a4a385f50f27c7894bce94a015e0303178fa. Please take a look when you have a chance :)

Issue #88, opened 2020-10-11
Getting ESLint errors about missing files


peterp:
I'm trying to add this plugin to my project, I followed the standard configuration in the documentation:

![image](https://user-images.githubusercontent.com/44849/95672804-f7ce4f80-0ba3-11eb-902a-aee50ea2d099.png)

But am getting the following error in VSCode's ESLint output:

```terminal
[Info  - 9:40:12 AM] ESLint server is starting
[Info  - 9:40:12 AM] ESLint server running in node v12.14.1
[Info  - 9:40:12 AM] ESLint server is running.
[Info  - 9:42:43 AM] ESLint library loaded from: /Users/peterp/x/redwoodjs/example-invoice/node_modules/eslint/lib/api.js
[Info  - 9:42:56 AM] ENOTDIR: not a directory, stat '/Users/peterp/x/redwoodjs/example-invoice/web/src/index.js/0_/Users/peterp/x/redwoodjs/example-invoice/web/src/index.js' Occurred while linting /Users/peterp/x/redwoodjs/example-invoice/web/src/index.js/0_/Users/peterp/x/redwoodjs/example-invoice/web/src/index.js:1
[Info  - 9:42:56 AM] ENOTDIR: not a directory, stat '/Users/peterp/x/redwoodjs/example-invoice/web/src/pages/InvoicePage/subcomponents/InvoiceCell/InvoiceCell.js/0_document.graphql' Occurred while linting /Users/peterp/x/redwoodjs/example-invoice/web/src/pages/InvoicePage/subcomponents/InvoiceCell/InvoiceCell.js/0_document.graphql:0
```

I've pushed up a branch here: https://github.com/redwoodjs/example-invoice/pull/39/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

dotansimha:
@peterp I think some rules/plugins doesn't play along with `processor` feature, probably because of the sub-paths that ESLint creates for those. 
I noticed that with `prettier/prettier` for example, do you use it? if so, can you try to disable it? 

```
      {
        "files": ["*.graphql"],
        "parser": "@graphql-eslint/eslint-plugin",
        "plugins": ["@graphql-eslint"],
         "rules": {
            "prettier/prettier": "off"
         }
      }
```

peterp:
That solved it! Thanks!

dotansimha:
Great :) Thank you for the feedback @peterp ! 
[...]

PR #394, opened 2021-04-22
fix: original file should not be considered as file block

JounQin:
Related #88

ESLint supports `text` directly, no need to use the hacky way. See https://github.com/eslint/eslint/blob/master/lib/linter/linter.js#L1298

Besides, I've fixed related `eslint-plugin-prettier`'s issue at https://github.com/prettier/eslint-plugin-prettier/pull/401

### DIFF
--- a/.changeset/empty-donuts-destroy.md
+++ b/.changeset/empty-donuts-destroy.md
@@ -1,0 +1,11 @@
+---
+"@graphql-eslint/eslint-plugin": patch
+---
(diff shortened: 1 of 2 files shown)


dotansimha:
Thank you @JounQin ! Can you please add a changeset file? (Run `yarn changeset` and it will guide you).

JounQin:
@dotansimha Done!

Issue #277, opened 2021-01-21
rule unique-type-names always shows error


lioneltay:
The rule requires you provide a schema to the plugin to work against and if it finds any types in your graphql files that are defined in the schema it will tell you it is a duplicate. However, the schema itself is generated from these type definitions, so every type becomes an error as long as you have used it to generate your schema (in the case of a json file), or have deployed the changes to an endpoint (in the case of providing a url as the schema).

I understand that eslint can only look at one file at a time? So I'm not sure how this can be improved... but it doesn't seem very useful as it is. Unless I am missing something...

dnalborczyk:
> I understand that eslint can only look at one file at a time? 

`eslint` might, but this plugin pulls in _all_ your schema files and creates an abstract syntax tree which every rule uses.

lioneltay:
Is the rule working as intended at the moment? I think the problem is that the plugin is not able to distinguish what is in the schema because it comes from other files, and what is coming from the current file it is linting. It seems to collect your entire schema, then use that to determine whether or not a type already exists, but the schema itself has types from the current file it is linting, so it always results in it saying that the type already exists.

dnalborczyk:
@lioneltay just gave it a try myself, it seems this rule is broken.

B2o5T:
any news?

Urigo:
Thanks for the detailed reports!

I'm not adding a lot here but I'm just labeling it according to our new Contribution Guide and [issue flow](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow).

It looks like already got it into `stage 0`! thank you for that! 

Now in order to advance to `stage 1` we'll need a running reproduction.

After that maybe a failing test to advance it through the [stages](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow).

Thank you and sorry that this comment is not a complete solution (yet).

B2o5T:
@Urigo thank u ;) I added sample repo https://github.com/B2o5T/graphql-eslint-unique-type-names

Urigo:
that's awesome @B2o5T !

For [`stage 2`](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow), we'll need a failing test If someone wants to give it a go: https://github.com/dotansimha/graphql-eslint/tree/master/packages/plugin/tests

B2o5T:
[...]

PR #405, opened 2021-04-27
test for rule `unique-type-names` always shows error 

B2o5T:
test for issue: https://github.com/dotansimha/graphql-eslint/issues/277

### DIFF
--- a/.changeset/tidy-papayas-count.md
+++ b/.changeset/tidy-papayas-count.md
@@ -1,0 +1,5 @@
+---
+'@graphql-eslint/eslint-plugin': patch
+---
+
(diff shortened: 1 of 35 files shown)


dotansimha:
I'm not sure if we have a way to solve that issue since it's a rule coming from `graphql-js`, and it basically checks a type against a schema, and not with a schema. So we can't really tell. 
I think the best way to solve that is to ignore the schema for this specific rule, this way we can use `validateSDL` instead of `validate` and allow the validation to run for the specific file, without the loaded schema. 

I fixed that and pushed it to this PR, seems to work now :)

Issue #406, opened 2021-04-27
Rule `lone-schema-definition` is broken


B2o5T:
It's shows errors for one schema definition and not show expected  error `Must provide only one schema definition.` for two schema definitions

B2o5T:
I skipped `stage-0` and added tests in order to advance to `stage 1`

dotansimha:
Fixed in `v1.0.0`! 🎉

PR #407, opened 2021-04-27
test for `lone-schema-definition` rule

B2o5T:
test for issue: https://github.com/dotansimha/graphql-eslint/issues/406


dotansimha:
Fixed 🎉 Like we had with `unique-type-name`, this should use `validateSDL`

PR #431, opened 2021-05-18
Fix link on unique-type-names.md

ardatan:


dotansimha:
@ardatan merged, but please note that this will break soon because of this `.md` is generated :D

ardatan:
@dotansimha Oh really 🤦‍♂️

Issue #385, opened 2021-04-13
Support #import


dobesv:
Our graphql systems support an `#import` directive in graphql files to allow sharing fragments between operations in different files.

e.g.

```
#import './TeacherProfileUser.fragment.graphql'
#import './ProfileSettingsContainer/ProfileSettings/StandardsPreferences/StandardsPreferences.fragment.graphql'
#import "../../core/viewer.fragment.graphql"

mutation TeacherProfileUpdateUser($userId: ID!, $input: UserInput!) {
  updateUser(id: $userId, input: $input) {
    user {
      ...TeacherProfileUserFragment
      ...StandardsPreferencesUserFragment
      ...ViewerFragment
    }
  }
}

```

Currently this file gives an error:

```
/home/ubuntu/app/src/ui/client/teacher/profile/TeacherProfile.mutation.graphql
   8:12  error  Unknown fragment "TeacherProfileUserFragment"        @graphql-eslint/known-fragment-names
   9:12  error  Unknown fragment "StandardsPreferencesUserFragment"  @graphql-eslint/known-fragment-names
  10:12  error  Unknown fragment "ViewerFragment"                    @graphql-eslint/known-fragment-names
```

It would be nice if this support could be ported over to this linter somehow.

oliverlukesch:
Stumbled upon the same problem, support for imports would be greatly appreciated

ardatan:
@oliverlukesch For now you can provide your operations under `parserOptions` like below;
```json
{
  "parserOptions": {
     "operations": "/path/to/operations/*.graphql"
   }
}
```

PR #441, opened 2021-05-19
🎉 New feature: Support `#import` for `known-fragment-names` rule

B2o5T:
Finally I made it work. Any feedback is welcome
Fixes https://github.com/dotansimha/graphql-eslint/issues/385

## Type of change

## Checklist:

### DIFF
--- a/.changeset/dry-vans-smile.md
+++ b/.changeset/dry-vans-smile.md
@@ -1,0 +1,5 @@
+---
+'@graphql-eslint/eslint-plugin': minor
+---
+
+support #import in known-fragment-names rule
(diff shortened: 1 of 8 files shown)


dotansimha:
Nice! Thank you @B2o5T ! 
We are using `graphql-tools` under the hood, and it already has support for `#import` syntax, so maybe we can use that? i guess instead of regular `parse`.  Maybe @ardatan can help here? :)

B2o5T:
@dotansimha a little update of my PR I made refactoring for using `siblings.getFragmentsInUse` and I added a new option for this rule `allowImportFragments`.
But now there is one case when rule will work not as expected.

```gql
#import './NON_EXISTENT_IMPORT.graphql'

query {
  user {
    ...UserFields
  }
}
```

For this file if it import file that don't exist, but if we loaded some fragments with name `UserFields` via `parserOptions.operations`, there will be no report of error

B2o5T:
@dotansimha currently in `@graphql/eslint` parser we parse graphql ast using `parseGraphQLSDL` from `@graphql-tools/utils` that don't support `#import` syntax.

I made changes in parser. For non virtual files I parse graphql ast with `GraphQLFileLoader` from `@graphql-tools/graphql-file-loader` that allows import and export definitions via `#import` syntax.

dotansimha (approved):

PR #451, opened 2021-05-25
🎉 New rule `no-unused-fields`

B2o5T:
## Description

Report about unused fields based on siblings operations/fragments/inline fragments

Fixes https://github.com/dotansimha/graphql-eslint/issues/9

## Type of change

## Checklist:

### DIFF
--- a/packages/plugin/src/graphql-ast.ts
+++ b/packages/plugin/src/graphql-ast.ts
@@ -16,7 +16,9 @@ import {
   isListType,
   isNonNullType,
   GraphQLDirective,
+  Kind,
 } from 'graphql';
+import { SiblingOperations } from './sibling-operations';
 
 export function createReachableTypesService(schema: GraphQLSchema): () => Set<string>;
 export function createReachableTypesService(schema?: GraphQLSchema): () => Set<string> | null {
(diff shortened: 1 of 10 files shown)


packages/plugin/src/graphql-ast.ts
```diff
@@ -128,3 +130,87 @@ export function collectReachableTypes(schema: GraphQLSchema): Set<string> {
...
+
+      if (node.kind === Kind.FIELD) {
+        const fieldName = node.name.value;
+        addField(type.name, fieldName);
+
+        if (node.selectionSet) {
+          let parentType = (type.astNode as any).fields.find(n => n.name.value === fieldName);
+          // Skipping if not found in schema
+          if (parentType) {
+            while (parentType.type) {
```
dotansimha, line 196: We have a helper called `getBaseType` for that logic

packages/plugin/src/graphql-ast.ts
```diff
@@ -128,3 +130,87 @@ export function collectReachableTypes(schema: GraphQLSchema): Set<string> {
...
+  };
+
+  const visit = (selection, type?: GraphQLNamedType) => {
+    switch (selection.kind) {
+      case Kind.OPERATION_DEFINITION:
+        type = typesByOperations[selection.operation];
+        break;
+      case Kind.FRAGMENT_DEFINITION:
+      case Kind.INLINE_FRAGMENT:
+        type = schema.getType(selection.typeCondition.name.value);
```
dotansimha, line 173: I think `typeCondition` could be `null`, we had an issue with that in codegen, worth adding the extra check.

packages/plugin/src/graphql-ast.ts
```diff
@@ -128,3 +130,87 @@ export function collectReachableTypes(schema: GraphQLSchema): Set<string> {
...
+            }
+            const parent = schema.getType(parentType.name.value);
+            visit(node, parent);
+          }
+        }
+      }
+    }
+  };
+
+  for (const { document } of operations.getFragments()) {
```
dotansimha, line 207: I think this could be changed with just a `visit` call using GraphQL `Visitor`, this way you don't need to write the recursive code that handles all edge cases. 
You can do:

```
visit(allDocuments, {
   Field: node => {
        // To get the type, you can use `typeInfo` here, no need to access schema directly,
        // you also can just access the field as-is, without the need to maintain logic that iterates recursively over the selection sets 
   }
});

dotansimha (commented):
This is awesome, thank you @B2o5T ! 
I added a few comments, let me know what you think :)

dotansimha:
Oh, and I think you might be able to use `graphql-inspector`'s `coverage` command as reference, @kamilkisiela what do you think?

packages/plugin/src/graphql-ast.ts
B2o5T: Thank @dotansimha , but after refactor in favor of using `visit` and `visitWithTypeInfo` from `graphql-js` I think I don't need anymore this helper

packages/plugin/src/graphql-ast.ts
B2o5T: Same, after refactor I don't use this code anymore, but thanks 👍

B2o5T:
> Oh, and I think you might be able to use `graphql-inspector`'s `coverage` command as reference, @kamilkisiela what do you think?

Thank you @dotansimha I looked source code of  `graphql-inspector`'s `coverage` and I make refactoring to use `visit` with `visitWithTypeInfo`

dotansimha:
Amazing work. Thank you @B2o5T ❤️

PR #487, opened 2021-06-09
ESM support

PabloSzx:
## Description

This is also a fix, since https://github.com/dotansimha/graphql-eslint/pull/486 was already merged and currently the project can't build without these changes

ESM Support

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

**Test Environment**:
- OS: Windows & macOS
- `@graphql-codegen/...`: latest 
- NodeJS: v14.17.0 & v16.3.0

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/.changeset/gorgeous-kids-dress.md
+++ b/.changeset/gorgeous-kids-dress.md
@@ -1,0 +1,5 @@
+---
+'@graphql-eslint/eslint-plugin': minor
+---
+
+ESM support
--- a/packages/plugin/package.json
+++ b/packages/plugin/package.json
@@ -5,9 +5,19 @@
   "author": "Dotan Simha <dotansimha@gmail.com>",
   "license": "MIT",
   "sideEffects": false,
-  "main": "dist/index.cjs.js",
-  "module": "dist/index.esm.js",
+  "main": "dist/index.js",
+  "module": "dist/index.mjs",
   "typings": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "require": "./dist/index.js",
+      "import": "./dist/index.mjs"
+    },
+    "./*": {
+      "require": "./dist/*.js",
+      "import": "./dist/*.mjs"
+    }
+  },
   "typescript": {
     "definition": "dist/index.d.ts"
   },
--- a/packages/plugin/src/utils.ts
+++ b/packages/plugin/src/utils.ts
@@ -2,6 +2,7 @@ import { Source, Lexer, GraphQLSchema, Token, DocumentNode } from 'graphql';
 import { GraphQLESlintRuleContext } from './types';
 import { AST } from 'eslint';
 import { SiblingOperations } from './sibling-operations';
+import * as gqlLanguage from 'graphql/language';
 
 export function requireSiblingsOperations(ruleName: string, context: GraphQLESlintRuleContext<any>): SiblingOperations {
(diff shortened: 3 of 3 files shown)


PabloSzx:
I am thinking that doesn't make sense for graphql-eslint to support ESM if eslint itself is all commonjs, and there is a couple places that the translation for esm is not straight forward:

https://github.com/dotansimha/graphql-eslint/blob/master/packages/plugin/src/testkit.ts#L26

https://github.com/dotansimha/graphql-eslint/blob/master/packages/plugin/src/rules/graphql-js-validation.ts#L56

dotansimha:
> I am thinking that doesn't make sense for graphql-eslint to support ESM if eslint itself is all commonjs, and there is a couple places that the translation for esm is not straight forward:
> 
> [https://github.com/dotansimha/graphql-eslint/blob/master/packages/plugin/src/testkit.ts#L26](https://github.com/dotansimha/graphql-eslint/blob/master/packages/plugin/src/testkit.ts?rgh-link-date=2021-06-09T17%3A34%3A46Z#L26)
> 
> [https://github.com/dotansimha/graphql-eslint/blob/master/packages/plugin/src/rules/graphql-js-validation.ts#L56](https://github.com/dotansimha/graphql-eslint/blob/master/packages/plugin/src/rules/graphql-js-validation.ts?rgh-link-date=2021-06-09T17%3A34%3A46Z#L56) does eslint support async validation to add inline dynamic imports?
> 
> In that case the solution for the current build error is to rollback the bob-the-bundler version, and add to bob-the-bundler an option to skip all the esm validation and building
> 
> PS: #486 shouldn't have been merged, the CI didn't catch the building error because "yarn build" it's not in the github actions workflow

Seems like ESLint does support ESM in some cases (like config file, see https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats ) . Not sure about parser/plugin and the rest of the features we are using. Since they are being loaded eventually by ESLint core, and we should expose code that is compatible with that. 

I agree that we can rollback Bob and then think of the next steps of this. What are you thoughts? @PabloSzx

PabloSzx:
> I agree that we can rollback Bob and then think of the next steps of this. What are you thoughts? @PabloSzx

Since bob-the-bundler version was already rolled-back the next steps are allowing the new bob-the-bundler to skip esm building, I did a quick read of eslint core and everything is plain js with commonjs, and is unlikely that eslint will do to change to esm in the short therm considering all the current ecosystem around it, so I would simply close this issue and add the bob-the-bundler configuration to commonjs only

dotansimha:
> > I agree that we can rollback Bob and then think of the next steps of this. What are you thoughts? @PabloSzx
> 
> Since bob-the-bundler version was already rolled-back the next steps are allowing bob-the-bundler to skip esm building, I did a quick read of eslint core and everything is plain js with commonjs, and is unlikely that eslint will do to change to esm in the short therm considering all the current ecosystem around it, so I would simply close this issue and add the bob-the-bundler configuration to commonjs only

Sounds good, I agree.

Issue #504, opened 2021-06-27
ESLint directives no longer working on nonvirtual files


B2o5T:
**Describe the bug**
ESLint directives no longer working on nonvirtual files

**To Reproduce**
Since we have fixed a bug to support `#import` comments, now nonvirtual files we parse with `GraphQLFileLoader` instead of `parseGraphQLSDL`, but after parsing `rawSDL` with `GraphQLFileLoader` all comments are trimmed.

**Environment:**

- OS: macOS
- `@graphql-eslint/...`: 1.1.0
- NodeJS: 14.16.0

ardatan:
@B2o5T Can we create a failing test on graphql-tools?

B2o5T:
@ardatan Done 🙂 Let me know what do you think

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@1.1.1`

PR #508, opened 2021-06-30
1️⃣ fix: make working ESLint directive in files that contains import comments

B2o5T:
Fixes https://github.com/dotansimha/graphql-eslint/issues/504

### DIFF
--- a/packages/plugin/src/rules/graphql-js-validation.ts
+++ b/packages/plugin/src/rules/graphql-js-validation.ts
@@ -1,8 +1,9 @@
-import { GraphQLESLintRule, GraphQLESlintRuleContext } from '../types';
-import { requireGraphQLSchemaFromContext } from '../utils';
-
 import { validate, GraphQLSchema, DocumentNode, ASTNode, ValidationRule } from 'graphql';
 import { validateSDL } from 'graphql/validation/validate';
+import { GraphQLFileLoader } from '@graphql-tools/graphql-file-loader';
+import fs from 'fs';
+import { GraphQLESLintRule, GraphQLESlintRuleContext } from '../types';
+import { requireGraphQLSchemaFromContext } from '../utils';
 import { GraphQLESTreeNode } from '../estree-parser';
 
 function extractRuleName(stack: string | undefined): string | null {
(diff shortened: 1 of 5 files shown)


packages/plugin/tests/eslint-directives.spec.ts
```diff
@@ -44,12 +45,8 @@ ruleTester.runGraphQLTests('test-directives', rule, {
       `,
     },
     {
-      code: /* GraphQL */ `
-        # eslint-disable
```
B2o5T, line 48: I deleted this test because he is same as test in line 40

packages/plugin/src/rules/graphql-js-validation.ts
```diff
@@ -62,35 +63,57 @@ const validationToRule = (
...
-            meta.docs.description +
+            docs.description +
             `\n\n> This rule is a wrapper around a \`graphql-js\` validation function. [You can find it's source code here](https://github.com/graphql/graphql-js/blob/main/src/validation/rules/${ruleName}Rule.ts).`,
         },
       },
-      create: context => {
+      create(context) {
+        let documentNode: DocumentNode;
+
+        const isGraphQLImportFile = rawSDL => {
```
B2o5T, line 85: make it's better export and import this function ? https://github.com/ardatan/graphql-tools/blob/master/packages/loaders/graphql-file/src/index.ts#L35

packages/plugin/src/rules/graphql-js-validation.ts
```diff
@@ -44,11 +45,16 @@ export function validateDoc(
   }
 }
 
+const isGraphQLImportFile = rawSDL => {
```
B2o5T, line 48: maybe it's better to export and import this function? https://github.com/ardatan/graphql-tools/blob/master/packages/loaders/graphql-file/src/index.ts#L35

dotansimha: Yeah, I agree, but it shouldn't be a blocker for this PR. @ardatan can you please export this function and release a new version?

dotansimha (approved):

Issue #517, opened 2021-07-05
Cannot read property 'length' of undefined (v. 1.1.1)


tobiasdiez:
**Describe the bug**

Upgrading from 1.1.0 to 1.1.1 leads to the following error for each graphql file.
```
Parsing error: Cannot read property 'length' of undefined
```

**To Reproduce**
I don't know a minimal config to reproduce this, but upgrading triggered this. See https://github.com/JabRef/JabRefOnline/pull/234.

**Expected behavior**

No error.

**Environment:**

- OS: Ubuntu
- `@graphql-eslint/...`: 1.1.1, 1.1.0 was working properly
- NodeJS: 14

**Additional context**

After a quick scroll over the changes, maybe `documents` here is null?
https://github.com/dotansimha/graphql-eslint/compare/@graphql-eslint/eslint-plugin@1.1.0...@graphql-eslint/eslint-plugin@1.1.1#diff-e211fbc78585e02682a66a6ac90f5bc5ce71748f3e1028f17bb75fc8511a7ccdR73

cvng:
@dotansimha I can confirm this is fixed by adding a  blank `documents` section in the config file

B2o5T:
Yes, `documents` can be undefined, I added a fix

B2o5T:
> @B2o5T any idea why this could happen?

Before we used `gqlConfig.getDocumentsSync`, and now `gqlConfig.loadDocumentsSync` that allow passing option`skipGraphQLImport`, no need to check `documents.length`, only need check to truthy of `documents`

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@1.1.2`. Thanks @tobiasdiez @B2o5T

PR #518, opened 2021-07-05
fix: documents in graphql-config can be null

B2o5T:
Fixes https://github.com/dotansimha/graphql-eslint/issues/517


tobiasdiez:
Thanks for the quick fix!

Issue #526, opened 2021-07-08
Cannot query field "admin_notification" on type "Subscription"  when multiple projects defined in .graphqlrc.yml


waterdrop01:
**Describe the bug**

Here is my `.graphqlrc.yml`:
```
projects:
  first:
    schema: ./first.graphql
    documents: ./src/first-gql.js
  second:
    schema: ./second.graphql
    documents: ./src/second-gql.js
```

Here is my `./src/first-gql.js` file:
```
const queryUsers = gql`
  query users($first: Int) {
    users(first: $first) {
      id
      symbol
    }
  }
`;
```

Here is my `./src/second-gql.js` file:
```
const subscribeAdminNotification = gql`
  subscription admin_notification(
    $limit: Int
    $order_by: [admin_notification_order_by!]
    $where: admin_notification_bool_exp
  ) {
    admin_notification(limit: $limit, order_by: $order_by, where: $where) {
      id
      level
    }
  }
`;
```
When running `npx eslint --ext graphql,js .` the `graphql-eslint` linter complains that `Cannot query field "admin_notification" on type "Subscription"`, because it is looking in the wrong `.graphql` schema!!!

If I remove the first schema definition from `.graphqlrc.yml`, then it works OK.

**Expected behavior**

`graphql-eslint` should lint .js files against their related .graphql schema definition (as defined by the `documents` field in `.graphqlrc.yml`)

Otherwise, it breaks compatibility with `graphql-config@3` which documentation states that [it's supporting multiple projects](https://graphql-config.com/usage#projects)

**May I know how do you lint when you have multiple .graphql files/projects in `.graphqlrc.yml`?**

Thanks for your help!

B2o5T:
Hi @waterdrop01 thank you for the detailed reports!
@dotansimha I think currently it's not possible to handle multischemas projects?

waterdrop01:
Here is the [`graphql-config@3` documentation that is referring to multiple projects support](https://graphql-config.com/usage#projects)

dotansimha:
I think multi-project should work, since we are using `graphql-config`, and it does support. 
We are using `getProjectForFile` to get the correct schema per each file we are running `graphql-eslint` on, so if the file is included correctly in the project of `graphql-config`, it should load the schema correctly. 

[...]

PR #532, opened 2021-07-12
fix(graphql-config): pass real filepath to `gqlConfig.getProjectForFile`

B2o5T:
Fixes https://github.com/dotansimha/graphql-eslint/issues/526

### DIFF
--- a/packages/plugin/src/schema.ts
+++ b/packages/plugin/src/schema.ts
@@ -8,6 +8,7 @@ import { buildSchema, GraphQLSchema } from 'graphql';
 import { GraphQLConfig } from 'graphql-config';
 import { dirname } from 'path';
 import { ParserOptions } from './types';
+import { getOnDiskFilepath } from './utils';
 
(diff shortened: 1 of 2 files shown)


packages/plugin/src/schema.ts
```diff
@@ -32,25 +33,23 @@ export function getSchema(options: ParserOptions, gqlConfig: GraphQLConfig): Gra
...
+  if (gqlConfig && options.filePath) {
+    const realFilepath = getOnDiskFilepath(options.filePath);
+    const fileDir = dirname(realFilepath);
 
     if (schemaCache.has(fileDir)) {
       schema = schemaCache.get(fileDir);
     } else {
-      const projectForFile = gqlConfig.getProjectForFile(options.filePath);
-
-      if (projectForFile) {
```
B2o5T, line 43: `projectForFile` is always truthy based on type definition. And I think this will be great if we add` "strict": true` to our `tsconfig.json`.

Issue #515, opened 2021-07-02
Incorrect behaviour with operations in .js/.ts files


mikew:
**Describe the bug**
The `@graphql-eslint/unique-fragment-name`, `@graphql-eslint/unique-operation-name`, and `@graphql-eslint/known-fragment-names` rules seem to incorrectly report errors when operations are defined in .js/.ts files.

**To Reproduce**

https://github.com/mikew/graphql-eslint-woes

**Expected behavior**
There should be no issues in the repo above.

**Environment:**

- OS: macOS
- `@graphql-eslint/...`: latest
- NodeJS: 14

dotansimha:
Hi @mikew , thank you for reporting this!
I think @B2o5T's changes in https://github.com/dotansimha/graphql-eslint/pull/511 might fix this?

B2o5T:
@dotansimha No :) it's another issue. Inside `unique-fragment-name` rule we make assertion
```js
normalizePath(f.filePath) !== normalizePath(context.getFilename()
```
where `context.getFilename()` is virtual file `/Users/dimitri/graphql-eslint-woes/src/index.js/0_document.graphql` and `f.filePath` is sibling with real file path `/Users/dimitri/graphql-eslint-woes/src/index.js`

mikew:
Thanks for looking into this so quickly! Curious, what about the `loadSiblings` snippet above seems strange?

dotansimha:
> Thanks for looking into this so quickly! Curious, what about the `loadSiblings` snippet above seems strange?

We have 3 different kinds of files loaded in this plugin:
1. Regular `.graphql` file - loaded with ESLint core using the plugin, and then we parse it with GraphQL and convert it to ESTree.
2. Code files with GraphQL code - we call these "virtual files" since we use ESLint processor to extract these from JS/TS files, and then we can a virtual file path (that doesn't exists in FS).
3. Siblings - these are operations loaded as "siblings" to either regular file or virtual files, and they are in use in order to support or enrich the plugins with more info (For example, you might use ESLint for lining a GraphQL file with a query, but if that query is using a fragment spread, we have a mechanism that loads the fragment)

The snippet @B2o5T mentioned is the one that compares file paths while loading siblings. When it's loaded against while linting a virtual file, we probably compare it in the wrong way (maybe we should use the original file path, not the virtual?).

B2o5T:
[...]

PR #535, opened 2021-07-13
4️⃣ fix false negative case in unique-fragment-name rule

B2o5T:
Related https://github.com/dotansimha/graphql-eslint/issues/515

`@graphql-tools/graphql-tag-pluck` v7 returns multiple extracted documents instead of single like in v6 so I updated `@graphql-tools` dependencies to v7 and I added support for virtual filepath for siblings

### DIFF
--- a/packages/plugin/src/sibling-operations.ts
+++ b/packages/plugin/src/sibling-operations.ts
@@ -1,3 +1,4 @@
+import { resolve } from 'path';
 import {
   FragmentDefinitionNode,
   FragmentSpreadNode,
@@ -29,6 +30,23 @@ export type SiblingOperations = {
   getOperationByType(operationType: OperationTypeNode): OperationSource[];
 };
 
+const handleVirtualPath = (documents: Source[]): Source[] => {
+  const filepathMap: Record<string, number> = {};
+
+  return documents.map(source => {
+    const { location } = source;
+    if (['.gql', '.graphql'].some(extension => location.endsWith(extension))) {
+      return source;
+    }
+    filepathMap[location] = filepathMap[location] ?? -1;
+    const index = filepathMap[location] += 1;
+    return {
+      ...source,
+      location: resolve(location, `${index}_document.graphql`)
+    };
+  });
+};
+
 const operationsCache: Map<string, Source[]> = new Map();
 const siblingOperationsCache: Map<Source[], SiblingOperations> = new Map();
 
@@ -56,7 +74,7 @@ const getSiblings = (filePath: string, gqlConfig: GraphQLConfig): Source[] => {
 };
 
 export function getSiblingOperations(options: ParserOptions, gqlConfig: GraphQLConfig): SiblingOperations {
-  const siblings = getSiblings(options.filePath, gqlConfig);
+  const siblings = handleVirtualPath(getSiblings(options.filePath, gqlConfig));
 
   if (siblings.length === 0) {
     let printed = false;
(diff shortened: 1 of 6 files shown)


packages/plugin/src/schema.ts
```diff
@@ -3,24 +3,15 @@ import { GraphQLFileLoader } from '@graphql-tools/graphql-file-loader';
...
+import { Loader } from '@graphql-tools/utils';
+import { GraphQLSchema } from 'graphql';
 import { GraphQLConfig } from 'graphql-config';
 import { dirname } from 'path';
 import { ParserOptions } from './types';
 
 const schemaCache: Map<string, GraphQLSchema> = new Map();
 
-export const schemaLoaders: Loader<string, SingleFileOptions>[] = [
-  {
```
B2o5T, line 15: did we need it before? I think it's incompatible with `graphql-tools` v7

packages/plugin/src/schema.ts
```diff
@@ -3,24 +3,15 @@ import { GraphQLFileLoader } from '@graphql-tools/graphql-file-loader';
...
-    loaderId: () => 'direct-string',
-    canLoad: async () => false,
-    load: async () => null,
-    canLoadSync: pointer => typeof pointer === 'string' && pointer.includes('type '),
-    loadSync: pointer => ({
-      schema: buildSchema(pointer),
```
B2o5T, line 21: did we need it before? I think it's incompatible with `graphql-tools` v7

packages/plugin/src/sibling-operations.ts
```diff
@@ -1,35 +1,25 @@
...
+export const operationsLoaders: Loader[] = [
   new GraphQLFileLoader(),
   new CodeFileLoader(),
-  {
-    loaderId: () => 'direct-string',
-    canLoad: async () => false,
-    load: async () => null,
-    canLoadSync: pointer => typeof pointer === 'string' && pointer.includes('type '),
-    loadSync: pointer => ({
-      document: parse(pointer),
```
B2o5T, line 30: same, did we need it before?  I think it's incompatible with `graphql-tools` v7

dotansimha: I think you are right 🤔

packages/plugin/package.json
```diff
@@ -16,23 +16,22 @@
     "prepack": "bob prepack"
   },
   "dependencies": {
-    "@graphql-tools/code-file-loader": "~6.3.0",
-    "@graphql-tools/graphql-file-loader": "~6.2.0",
-    "@graphql-tools/graphql-tag-pluck": "~6.5.0",
+    "@graphql-tools/code-file-loader": "7.0.0-alpha-c42e811d.0",
```
dotansimha, line 19: Awesome! I'm not sure if we need to merge it with the new v7, because it's still moving fast. 
Maybe we can get a release for v6 of the loaders with the fixed that we need from there? @B2o5T  is it the comments fix only, or do we need something else?

B2o5T: I think we can't fix actual `code-file-loader` and `graphql-tag-pluck` because there is some breaking changes between v6 and v7 like `loadSync` of `CodeFileLoader` returns `Source[] | null` in v7 and `Source | null` in v6.

Or `gqlPluckFromCodeStringSync` of `graphql-tag-pluck` returns `Source[]` in v7 and `string` in v6
I think we need to wait for release `graphql-tools` v7. What do you think @dotansimha?

PS At least we already have PR that fix this issue 😅.

dotansimha:
@B2o5T I merged 1+2+3, can't merge this one because it's on your fork and I'm not able to rebase it. can you please check that? 

Thank you 😍

dotansimha (approved):

Issue #525, opened 2021-07-08
Unable to load schema from a URL


tubbo:
**Describe the bug**

When trying to load my schema from an external URL, I get an error coming from `node-fetch`:

```
TypeError: Expected signal to be an instanceof AbortSignal
    at new Request (/Users/tscott/Code/nerdst/node_modules/node-fetch/lib/index.js:1222:10)
    at Object.deserializeRequest (/Users/tscott/Code/nerdst/node_modules/sync-fetch/shared.js:45:10)
    at Socket.<anonymous> (/Users/tscott/Code/nerdst/node_modules/sync-fetch/worker.js:16:26)
    at Socket.emit (events.js:326:22)
    at endReadableNT (_stream_readable.js:1241:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

This happens on every URL that I throw at it

**To Reproduce**

Load your schema from a URL in GraphQL config:

```yaml
schema: http://localhost:8000/graphql
documents: src/**/*.gql
```

**Expected behavior**

The schema from the URL I provided should be used

**Environment:**

- OS: macOS v11.3, though I've also seen this happen in CI
- `@graphql-eslint/...`: `"^1.1.2"`
- NodeJS: v12.22.1

**Additional context**

I did some digging and noticed that `node-fetch` is somehow wrong about its assertion that the passed-in signal is not an AbortSignal. Something strange is going on with that logic and I'm pretty sure it's the root cause of all this, but plenty of other packages I depend on use the exact same `node-fetch` version without any problems. GraphQL Code Generator, using the exact same graphql-config (`.graphqlrc.yml`), can introspect the graph at this URL just fine, but as soon as I try to run ESLint, that's when things start to break down and I get that error all over the place.

Also wanted to say how helpful this project has been. `eslint-plugin-graphql` was what we were using before and honestly I just disabled it after a while because it didn't provide any useful feedback. This plugin, on the other hand, not only fully validates our operations but also our schema, which has been very helpful in identifying unused code, formatting mistakes, and lack of documentation. Thanks for making it!

B2o5T:
Hi @tubbo thank you very much! Your detailed report is very helpful! Your feedback is awesome ☺️.

I found where the issue comes, firstly from [@graphql-tools/url-loader](https://github.com/ardatan/graphql-tools/tree/master/packages/loaders/url) that uses [sync-fetch](https://www.npmjs.com/package/sync-fetch) under the hood.

[...]

PR #542, opened 2021-07-21
3️⃣ Test schema loaders, fix `Expected signal to be an instanceof AbortSignal` error when `UrlLoader` is used

B2o5T:
Related https://github.com/dotansimha/graphql-eslint/issues/525

To fix error we need to update `url-loader` to alpha version

### DIFF
--- a/.changeset/wild-rules-guess.md
+++ b/.changeset/wild-rules-guess.md
@@ -1,0 +1,5 @@
+---
+'@graphql-eslint/eslint-plugin': patch
+---
+
+add tests for schema loaders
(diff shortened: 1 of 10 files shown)


dotansimha:
Thanks, @B2o5T , that's awesome. I do want to wait a bit longer until @ardatan will be available and will release the new version of graphql-tools, and then we'll update it here.

B2o5T:
@dotansimha It seems like loading schema by URL works after refactoring for using `graphql-config` for load schema and siblings.
I think this is because in `graphql-config`  `url-loader` has version `^6.0.0` that don't have error `Expected signal to be an instanceof AbortSignal` when `sync-fetch` is used.

PR #564, opened 2021-08-03
enhance(eslint-plugin): refactor the parts using tools

ardatan:
This PR also adds `loaderCache` that removes the need of `schemaCache` and `operationsCache` but I didn't remove it in this PR because it seems to need more changes than removing them.

### DIFF
--- a/packages/plugin/src/graphql-ast.ts
+++ b/packages/plugin/src/graphql-ast.ts
@@ -70,7 +70,7 @@ export function getUsedFields(schema: GraphQLSchema, operations: SiblingOperatio
   if (process.env.NODE_ENV !== 'test' && usedFieldsCache) {
     return usedFieldsCache;
   }
-  const usedFields: UsedFields = {};
+  const usedFields: UsedFields = Object.create(null);
   const typeInfo = new TypeInfo(schema);
   const allDocuments = [...operations.getOperations(), ...operations.getFragments()];
 
--- a/packages/plugin/src/graphql-ast.ts
+++ b/packages/plugin/src/graphql-ast.ts
@@ -1,5 +1,5 @@
-import { GraphQLSchema, TypeInfo, ASTKindToNode, Visitor, visit, visitWithTypeInfo, parse } from 'graphql';
-import { printSchemaWithDirectives } from '@graphql-tools/utils';
+import { GraphQLSchema, TypeInfo, ASTKindToNode, Visitor, visit, visitWithTypeInfo } from 'graphql';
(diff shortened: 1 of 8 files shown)


packages/plugin/src/schema.ts
```diff
@@ -17,12 +17,15 @@ export function getSchema(options: ParserOptions = {}, gqlConfig: GraphQLConfig)
     return null;
   }
 
-  if (schemaCache.has(schemaKey)) {
```
ardatan, line 20: Using `get` once will be better for performance

packages/plugin/src/sibling-operations.ts
```diff
@@ -61,15 +61,16 @@ const getSiblings = (filePath: string, gqlConfig: GraphQLConfig): Source[] => {
...
 
-  const documents = projectForFile.loadDocumentsSync(projectForFile.documents, {
-    skipGraphQLImport: true,
-  });
-  const siblings = handleVirtualPath(documents)
-  operationsCache.set(documentsKey, siblings);
+  if (!siblings) {
+    const documents = projectForFile.loadDocumentsSync(projectForFile.documents, {
+      skipGraphQLImport: true,
+      cache: loaderCache
```
ardatan, line 69: This might improve the performance a bit as well

packages/plugin/src/graphql-ast.ts
```diff
@@ -32,10 +31,12 @@ export function getReachableTypes(schema: GraphQLSchema): ReachableTypes {
       node.operationTypes.forEach(collect);
     },
     ObjectTypeDefinition(node) {
-      [node, ...node.interfaces].forEach(collect);
```
ardatan, line 35: This might cause multiple iterations

packages/plugin/src/graphql-ast.ts
```diff
@@ -12,10 +12,9 @@ export function getReachableTypes(schema: GraphQLSchema): ReachableTypes {
   if (process.env.NODE_ENV !== 'test' && reachableTypesCache) {
     return reachableTypesCache;
   }
-  // 👀 `printSchemaWithDirectives` keep all custom directives and `printSchema` from `graphql` not
-  const printedSchema = printSchemaWithDirectives(schema); // Returns a string representation of the schema
-  const astNode = parse(printedSchema); // Transforms the string into ASTNode
-  const cache = Object.create(null);
+
+  const astNode = getDocumentNodeFromSchema(schema); // Transforms the schema into ASTNode
```
ardatan, line 16: `printSchemaWithDirectives` also uses this function and no need to do `print` and `parse` again.

Issue #73, opened 2020-10-05
New rule: return a field of type Query in the mutation response


FluorescentHallucinogen:
See https://graphql-rules.com/rules/mutation-payload-query

PR #643, opened 2021-09-21
2️⃣ 🎉 feat(new rule): `require-field-of-type-query-in-mutation-result`

B2o5T:
related https://github.com/dotansimha/graphql-eslint/issues/73


packages/plugin/src/rules/require-deprecation-date.ts
```diff
@@ -14,7 +14,7 @@ const rule: GraphQLESLintRule<[{ argumentName?: string }]> = {
     docs: {
       category: 'Best Practices',
       description:
-        'Require deletion date on @deprecated directive. Suggest to remove deprecated things after deprecated date.',
+        'Require deletion date on `@deprecated` directive. Suggest removing deprecated things after deprecated date.',
```
B2o5T, line 17: `Suggest removing` is more correct

Issue #62, opened 2020-10-05
Better `options` API for ast-specific rules 


dotansimha:
Related: https://github.com/ilyavolodin/graphql-eslint/issues/3

PR #662, opened 2021-10-02
[breaking change] 🎉 `graphql-eslint` v3

B2o5T:
fixes https://github.com/dotansimha/graphql-eslint/issues/62 https://github.com/dotansimha/graphql-eslint/issues/664

There are breaking changes because options are changed.
This needs to be merged on the next major release, I can write also a migration guide to new options API.

Also, I saw that rules `avoid-operation-name-prefix` and `no-operation-name-suffix` rules we can remove in the next major release because we can validate the same things with `naming-convention` rule

Recommended config for `naming-convention` rule now looks like:

```js
{
  types: 'PascalCase',
  fields: 'camelCase',
  overrides: {
    EnumValueDefinition: 'UPPER_CASE',
    OperationDefinition: {
      style: 'PascalCase',
      forbiddenPrefixes: ['Query', 'Mutation', 'Subscription', 'Get'],
      forbiddenSuffixes: ['Query', 'Mutation', 'Subscription'],
    },
    FragmentDefinition: {
      style: 'PascalCase',
      forbiddenPrefixes: ['Fragment'],
      forbiddenSuffixes: ['Fragment'],
    },
    'FieldDefinition[parent.name.value=Query]': {
      forbiddenPrefixes: ['query', 'get'],
      forbiddenSuffixes: ['Query'],
    },
    'FieldDefinition[parent.name.value=Mutation]': {
      forbiddenPrefixes: ['mutation'],
      forbiddenSuffixes: ['Mutation'],
    },
    'FieldDefinition[parent.name.value=Subscription]': {
      forbiddenPrefixes: ['subscription'],
      forbiddenSuffixes: ['Subscription'],
    },
  },
}
```


dotansimha (approved):
Looks nice. I think it needs a rebase? And is it still a draft? 

I think we can do a minor release with eslint 8 support, and then do a major with these changed (and the other breaking changes)

docs/rules/naming-convention.md
```diff
@@ -14,151 +14,120 @@ Require names to follow specified conventions.
...
+- `ObjectTypeDefinition`
+- `OperationDefinition`
+- `ScalarTypeDefinition`
+- `UnionTypeDefinition`
+- `VariableDefinition`
 
-The object must be one of the following types:
+> It's also possible to use a [`selector`](https://eslint.org/docs/developer-guide/selectors) that starts with `ASTNode` name
+>
+> Example: pattern property `FieldDefinition[parent.name.value=Query]` will match only fields for type `Query`
```
JustinTRoss, line 98: Where can I find info about what values would go in the selector? E.g. here, you used `parent.name.value`. How did you know to use that?

docs/rules/naming-convention.md
```diff
@@ -14,151 +14,120 @@ Require names to follow specified conventions.
...
-### `ObjectTypeDefinition`
+- `FieldDefinition`
+- `InputValueDefinition`
+- `VariableDefinition`
+- `Argument`
+- `DirectiveDefinition`
 
 The object must be one of the following types:
 
 * `asString`
```
JustinTRoss, line 66: I know these already exist. I found them unclear though. Any reason we are prefixing with `as` rather than just saying "String"? Is that an ESLint thing?

docs/rules/naming-convention.md
```diff
@@ -14,151 +14,120 @@ Require names to follow specified conventions.
...
+> It's also possible to use a [`selector`](https://eslint.org/docs/developer-guide/selectors) that starts with `ASTNode` name
+>
+> Example: pattern property `FieldDefinition[parent.name.value=Query]` will match only fields for type `Query`
 
-* `asString`
-* `asObject`
+Pattern properties of the `overrides` object:
 
-### `ScalarTypeDefinition`
+#### `^(Argument|DirectiveDefinition|EnumTypeDefinition|EnumValueDefinition|FieldDefinition|FragmentDefinition|InputObjectTypeDefinition|InputValueDefinition|InterfaceTypeDefinition|ObjectTypeDefinition|OperationDefinition|ScalarTypeDefinition|UnionTypeDefinition|VariableDefinition)(.+)?$`
```
JustinTRoss, line 102: It isn't apparent what this is meant to convey in the docs.

docs/rules/naming-convention.md
```diff
@@ -179,6 +149,7 @@ The object is an array with all elements of the type `string`.
 Additional restrictions:
 
 * Minimum items: `1`
+* Unique items: `true`
 
 ## Resources
```
JustinTRoss, line 155: Consider adding in a suggested rule config folks can just paste in as a starting point. What you use in `recommended` is probably sufficient.

.changeset/chatty-shoes-hunt.md
```diff
@@ -0,0 +1,9 @@
+---
+'@graphql-eslint/eslint-plugin': major
+---
+
+feat: split `recommended` config to two configs `schema-recommended` and `operations-recommended`
+
+`recommended` and `all` configs were divided to 4 configs `schema-recommended`
+, `operations-recommended`, `schema-all` and `operations-all`. This was done in order to
+use `recommended` and `all` configs in `schema-only` projects where it is not possible to provide operations.
```
JustinTRoss, line 9: Per @B2o5T's feedback [here](https://github.com/dotansimha/graphql-eslint/pull/752#discussion_r744166414), `schema-all` is intended to assign all schema rules regardless of requirement for `parserOptions.operations`.

B2o5T: Hi @JustinTRoss, I changed this after your feedback, so now schema-only projects will be able to use schema-all config :)

docs/rules/naming-convention.md
B2o5T: Changed, what do think if now it's more clear?

docs/rules/naming-convention.md
B2o5T: Added more info 👍

packages/plugin/src/configs/schema-recommended.ts
```diff
@@ -0,0 +1,48 @@
+/*
+ * 🚨 IMPORTANT! Do not manually modify this file. Run: `yarn generate-configs`
+ */
+
+export default {
+  extends: ['plugin:@graphql-eslint/base'],
+  rules: {
+    '@graphql-eslint/description-style': 'error',
```
JustinTRoss, line 8: Consider including `require-description` in recommended set. It doesn't require any parser options, and it's definitely a widely recommended practice.

.changeset/three-shoes-relax.md
```diff
@@ -0,0 +1,61 @@
...
+    "error",
+    {
+      "ObjectTypeDefinition": "PascalCase",
+      "InterfaceTypeDefinition": "PascalCase",
+      "EnumTypeDefinition": "PascalCase",
+      "ScalarTypeDefinition": "PascalCase",
+      "InputObjectTypeDefinition": "PascalCase",
+      "UnionTypeDefinition": "PascalCase",
+      "FieldDefinition": "camelCase",
+      "InputValueDefinition": "camelCase",
```
JustinTRoss, line 41: Is this represented in the `after` set?

.changeset/three-shoes-relax.md
```diff
@@ -0,0 +1,61 @@
...
+After
+
+```json
+{
+  "@graphql-eslint/naming-convention": [
+    "error",
+    {
+      "types": "PascalCase",
+      "fields": "camelCase",
+      "FieldDefinition[parent.name.value=Query]": "camelCase"
```
JustinTRoss, line 57: Is this redundant, since fields are already set to `camelCase`?

README.md
```diff
@@ -184,11 +182,20 @@ You can find a complete list of [all available rules here](docs/README.md).
 
 ## Available Configs
 
-This plugin exports a [`recommended` config](packages/plugin/src/configs/recommended.ts) that enforces good practices and an [`all` config](packages/plugin/src/configs/all.ts) that makes use of all rules (except for deprecated ones).
+<!-- prettier-ignore-start -->
+|Name|Description|
+|:-:|-|
+|[`schema-recommended`](packages/plugin/src/configs/schema-recommended.ts)|enables all recommended rules|
+|[`operations-recommended`](packages/plugin/src/configs/operations-recommended.ts) |enables all recommended rules|
+|[`schema-all`](packages/plugin/src/configs/schema-all.ts)|enables all rules (except for the rules that require `parserOptions.operations` option)|
```
JustinTRoss, line 190:
```suggestion
|[`schema-all`](packages/plugin/src/configs/schema-all.ts)|enables all rules, except for those that require `parserOptions.operations` option.|
```

README.md
```diff
@@ -184,11 +182,20 @@ You can find a complete list of [all available rules here](docs/README.md).
...
+|[`operations-all`](packages/plugin/src/configs/operations-all.ts)|enables all rules|
```
JustinTRoss, line 191: Does this include rules that require schema `parserOption`?

.changeset/three-shoes-relax.md
B2o5T: Your are right, it’s redundant, it’s better to change to `forbiddenPrefixes: ['get']` and change in legacy `QueryDefinition` also. Nice catch 😀💪🏻

.changeset/polite-pillows-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-eslint/eslint-plugin': major
+---
+
+feat: `description-style` now have default description style `block`
```
JustinTRoss, line 5: Was this an arbitrary change from `inline` to `block`, or is there motivation behind it?

README.md
B2o5T: Yes, this include all operations rules, that require (or not) `parserOptions.schema` option

.changeset/polite-pillows-pump.md
B2o5T: GitHub GraphQL public schema has `block` comments, so I changed this to `block`, also with `block` comments you can have multiline comments but with `inline` not

README.md
B2o5T: Thanks, I’ll change this 🤗

docs/rules/naming-convention.md
JustinTRoss: Ahh. Yes. I understand what you were doing now. You were communicating in regexp? 😄 

It does seem a shame to need to have 10 sections all with just a couple words changes, but not sure there's any alternative 🤷‍♂️

B2o5T: Yes this was a regex, we use a package that generate markdown documentation based on json schema 🙂

.changeset/polite-pillows-pump.md
JustinTRoss: Got it. Thus, requiring `inline` would be disallowing multiline comments.  Thanks for the details. 🙏

.changeset/three-shoes-relax.md
B2o5T: yes, `fields` option include `InputValueDefinition` :)
image

packages/plugin/src/configs/schema-recommended.ts
B2o5T: added

.changeset/three-shoes-relax.md
JustinTRoss: Neat. That could actually make sense. `DirectiveDefinition` seems out of place here though.

docs/rules/naming-convention.md
B2o5T: Added more examples

.changeset/three-shoes-relax.md
B2o5T: I took proposition by @ilyavolodin https://github.com/ilyavolodin/graphql-eslint/issues/3#issuecomment-652155524
You are right that it's out of place in terms of the name of the group :)

JustinTRoss: I'm also a little surprised on the application of naming requirements on `Arguments`. If I understand correctly, `Arguments` are key/value pairs provided in operations, the keys of which must match the schema your operation is intended to consume, which you may not have control over. Thus, imposing naming requirements on `Argument` would be imposing only a constraint on which APIs you can consume.

Does that sound accurate? Or am I missing something super obvious 😅 ?

B2o5T:
@JustinTRoss thanks a lot for the review and suggestions. You can test the next major version by installing alpha version `3.0.0-alpha-4613dfe.0`. Any feedback is welcome 😜

docs/rules/naming-convention.md
B2o5T: Nope, it's not an ESLint thing, it's just a name of subschema definition.

JustinTRoss:
> @JustinTRoss thanks a lot for the review and suggestions. You can test the next major version by installing alpha version 3.0.0-alpha-4613dfe.0. Any feedback is welcome 😜

Thank you for doing all the hard work! 💪 🧠 
Working on getting upgraded to that alpha version presently 😀

.changeset/three-shoes-relax.md
B2o5T: What a catch 😍, you are totally right 💯. This was my mistake, I removed `Argument` from `operations-recommended` config and added a relevant test to prevent this mistake in the future 🙏

JustinTRoss: Along that same recommendation from Ilya, I'm curious if it would make sense to remove `fields (boolean)` option from `naming-convention` or add it to `require-description` to start to standardize.

B2o5T: I agree to remove that option, because:

1. we will take away the confusing presence of `DirectiveDefinition` in `fields` group.
2. without `DirectiveDefinition` in `fields` group we'll have only 3 kinds in `fields` group, so this group is really small in relation to `types` group where there are 6 kinds

JustinTRoss: That makes most sense to me as well. We already addressed above that naming-convention isn't helpful for `Argument` either, which effectively reduces this to two types, one of which wasn't apparent to me was included under the `fields` naming. I think removing it makes the API easier to use and harder to misuse.

packages/plugin/src/configs/schema-all.ts
```diff
@@ -0,0 +1,23 @@
...
+    '@graphql-eslint/alphabetize': [
+      'error',
+      {
+        fields: ['ObjectTypeDefinition', 'InterfaceTypeDefinition', 'InputObjectTypeDefinition'],
+        values: ['EnumTypeDefinition'],
+        arguments: ['FieldDefinition', 'Field', 'DirectiveDefinition', 'Directive'],
+      },
+    ],
+    '@graphql-eslint/input-name': 'error',
+    '@graphql-eslint/no-root-type': 'off',
```
JustinTRoss, line 17: Why setting these explicitly to 'off'? I'm having trouble seeing a scenario where folks would want extending their config with this ruleset to turn off these rules for them.

B2o5T: This rule requires a provided option, I have set to `off` to explicitly show which rules are disabled in the `all` config, if the user will want to enable this rule it's totally fine

.changeset/three-shoes-relax.md
B2o5T: `naming-convention` is helpful for `Argument` only in the schema but not in operations. `fields` option was removed 🙂

B2o5T:
@JustinTRoss do you have a twitter account?

dotansimha (approved):

JustinTRoss:
> @JustinTRoss do you have a twitter account?
`@JustinOfRoss`

packages/plugin/src/configs/schema-all.ts
JustinTRoss: Got it. Sounds like it's a part of the way things are generated.

Low impact, but setting explicitly to 'off' may inadvertently place requirements on folks' extends chains. E.g.:

```
    ...
    extends: [
        'prettier/@typescript-eslint',
        'plugin:@justin-eslint',
        'plugin:@graphql-eslint/schema-all'
  ],
```

If `@justin-eslint` sets `no-root-type`, extending it before `@graphql-eslint/schema-all` will result in the rule being unset, which I think might be a surprise for folks.

PR #663, opened 2021-10-02
[breaking change] modify options API for `require-description` rule

B2o5T:
related https://github.com/dotansimha/graphql-eslint/issues/62

### DIFF
--- a/docs/rules/require-description.md
+++ b/docs/rules/require-description.md
@@ -12,7 +12,7 @@ Enforce descriptions in your type definitions.
 ### Incorrect
 
 ```graphql
-# eslint @graphql-eslint/require-description: ['error', { on: ['ObjectTypeDefinition', 'FieldDefinition'] }]
+# eslint @graphql-eslint/require-description: ['error', { types: true, overrides: { FieldDefinition: true } }]
 
 type someTypeName {
   name: String
@@ -22,7 +22,7 @@ type someTypeName {
 ### Correct
 
 ```graphql
-# eslint @graphql-eslint/require-description: ['error', { on: ['ObjectTypeDefinition', 'FieldDefinition'] }]
+# eslint @graphql-eslint/require-description: ['error', { types: true, overrides: { FieldDefinition: true } }]
 
 """
 Some type description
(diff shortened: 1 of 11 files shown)


packages/plugin/tests/require-description.spec.ts
```diff
@@ -100,10 +95,14 @@ ruleTester.runGraphQLTests('require-description', rule, {
...
-          on: [Kind.OBJECT_TYPE_DEFINITION, Kind.FIELD_DEFINITION],
+          types: true,
+          overrides: {
+            ObjectTypeDefinition: false,
+            FieldDefinition: true,
+          },
         },
       ],
-      errors: 3,
+      errors: 2,
```
B2o5T, line 105: because I tested `overrides` property by setting `ObjectTypeDefinition: false`

dotansimha:
@B2o5T what's the status of this one? is it a breaking changes for the settings?

docs/rules/require-description.md
```diff
@@ -39,13 +39,42 @@ type someTypeName {
...
+- `UnionTypeDefinition`
 
-* Minimum items: `1`
+### `overrides` (object)
+
+Configuration for precise `ASTNode`
+
+Properties of the `overrides` object:
+
+#### `ObjectTypeDefinition` (boolean)
```
JustinTRoss, line 59: My guess is that the definitions for each of these won't be self-evident for most folks. It may be worth including either a basic definition of what each is or a link to their reference in graphql spec:

https://spec.graphql.org/June2018/#ObjectTypeDefinition
https://spec.graphql.org/June2018/#InterfaceTypeDefinition
https://spec.graphql.org/June2018/#EnumTypeDefinition
https://spec.graphql.org/June2018/#ScalarTypeDefinition
https://spec.graphql.org/June2018/#InputObjectTypeDefinition
https://spec.graphql.org/June2018/#UnionTypeDefinition
https://spec.graphql.org/June2018/#FieldDefinition
https://spec.graphql.org/June2018/#InputValueDefinition
https://spec.graphql.org/June2018/#EnumValueDefinition
https://spec.graphql.org/June2018/#DirectiveDefinition

B2o5T: @JustinTRoss I like your proposal, how about this one https://github.com/dotansimha/graphql-eslint/blob/new-options-for-require-description/docs/rules/require-description.md and https://github.com/dotansimha/graphql-eslint/blob/new-options-for-require-description/docs/rules/naming-convention.md
Also, I referenced the last spec of `October2021` instead of `June2018` :)

JustinTRoss: Haha. Yes, definitely. Good catch. As I sent it, I was a little surprised that the spec hadn't been updated since June2018. 😅

PR #735, opened 2021-10-30
fix error report for `unique-fragment-name` and `unique-operation-name` rule

B2o5T:


packages/plugin/src/rules/unique-fragment-name.ts
```diff
@@ -14,11 +14,7 @@ export const checkNode = (
   ruleName: string,
   messageId: string
 ): void => {
-  const documentName = node.name?.value;
-  if (!documentName) {
```
B2o5T, line 18: I changed AST selector to `OperationDefinition[name!=undefined]` so this `if` no longer needed

dotansimha (approved):

dotansimha:
@B2o5T   Great! Only changeset is missing ;)

PR #746, opened 2021-10-31
fix error report for `require-deprecation-reason` and `require-field-of-type-query-in-mutation-result` rule

B2o5T:


packages/plugin/src/rules/avoid-duplicate-fields.ts
```diff
@@ -67,61 +56,51 @@ const rule: GraphQLESLintRule<[], false> = {
       ],
     },
     messages: {
-      [AVOID_DUPLICATE_FIELDS]: `{{ type }} "{{ fieldName }}" defined multiple times.`,
+      [AVOID_DUPLICATE_FIELDS]: `{{ type }} "{{ fieldName }}" defined multiple times`,
     },
     schema: [],
   },
   create(context) {
+    function checkNode(
```
B2o5T, line 64: I simplified a bit this rule, removed and reduced lines of code

dotansimha (approved):

PR #747, opened 2021-10-31
fix error report for `selection-set-depth` rule and for `graphql-js` rules

B2o5T:


packages/plugin/src/rules/no-anonymous-operations.ts
```diff
@@ -37,27 +38,14 @@ const rule: GraphQLESLintRule = {
...
-                line: start.line,
-              },
-            },
-            data: {
-              operation: node.operation,
-            },
-            messageId: NO_ANONYMOUS_OPERATIONS,
-          });
-        }
+      'OperationDefinition[name=undefined]'(node) {
```
B2o5T, line 41: I think now it's impossible to make this rule more simple 🥳😂

dotansimha: I love it!!!!

dotansimha (approved):

Issue #552, opened 2021-07-29
Evaluating recommended ruleset


mhatvan:
**Is your feature request related to a problem? Please describe.**

At the current project I am working on, we are in the process of evaluating a recommended ruleset for our usage of `graphql-eslint`, since there is none yet unfortunately. The project is quite big with 50+ packages in a monorepo structure, so this plugin will have a great impact (thanks for making it!)

Basically, around 5-7 people in our team will vote on keeping or removing rules for our ruleset and then we will agree on the final ruleset. I can post the resulting ruleset here, then it could be further discussed and potentially implemented in `graphql-eslint` as recommended to make devs work easier!

dotansimha:
Hi @mhatvan ! We'll be really happy to hear your thoughts and ideas! Thank you for this valuable feedback, and thank you for taking the time to open this issue! :) 

Please, feel free to share with us your set of recommended rules and configuration, and feel free to share any feedback regarding the library itself.

mhatvan:
Sure thing, the least I can do for using such a great library!

It will still take a couple of days until all relevant devs will decide on the ruleset, then I will let you know right away.

mhatvan:
It took a bit longer since we only have bi-weekly crew days, but here is the first rough draft:
https://gist.github.com/mhatvan/160b533836bcf9fa5ce01c9f47d101da

It's in a gist because it will still need some iterations until its finished.

This is how to read the ruleset:
- rules that are enabled and set to 1 are rules that had > 50% approval rating
- rules that are commented out and set to 1 are rules that had < 50% approval rating, might still be discussed internally
- rules that are set to 0 are rules that everyone voted to remove
- rules that are commented out with `TODO:` are wanted, but we are having issues with it currently

dotansimha:
Thanks @mhatvan !! 
Recently @B2o5T added some setup for the recommended rules-set, I think it's based on this list, right?

B2o5T:
@dotansimha at the moment I added to recommended config only all rules from `graphql-js` and the rules that already had `recommended: true`. Also, I found that some of our rules need to have an additional configuration like [naming-convention](https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/naming-convention.md#correct), so we need to think about

dotansimha:
Closing, the new rules set is available in `@graphql-eslint/eslint-plugin@3.0.0`

Issue #645, opened 2021-09-22
plugin:@graphql-eslint/all runs into error caused by incorrect "require-description" lint config


stephenkoo:
**Describe the bug**

When running eslint using `plugin:@graphql-eslint/all`, get the error:

```
Error: .eslintrc.js#overrides[1] » plugin:@graphql-eslint/all:
        Configuration for rule "@graphql-eslint/require-description" is invalid:
        Value [] should NOT have fewer than 1 items.
```

This is even when disabling `"@graphql-eslint/require-description"`:
```js
    {
      files: ["*.graphql"],
      extends: ["plugin:@graphql-eslint/all"],
      rules: {
        "@graphql-eslint/require-description": "off",
      },
    },
```

**To Reproduce**

Create an eslint.js as above and run eslint.

**Expected behavior**

Lint to run without error.

**Environment:**

- OS: macOS 11.6
- `@graphql-eslint/...`: 2.2.0
- NodeJS: 16.9.1

**Additional context**

I believe this issue has a similar fix to the ones applied here:
- https://github.com/LD4P/sinopia_editor/issuec/1276
- https://github.com/airbnb/javascript/issues/2105

JustinTRoss:
FYI - This issue is now reproducible in this repo 😅 :
https://github.com/JustinTRoss/graphql-eslint-issue-repro

B2o5T:
Yes, I fixed this issue but only for next major version of graphql-eslint, that will be released very soon (we splitted `recommended` and `all` config into 4 configs - `schema-recommended`, `operations-recommended`, `schema-all` and `operations-all` configs)

`graphql-eslint` v3 is coming 🤟

lughino:
Hi @B2o5T , is there an ETA for releasing v3?
I'm currently blocked from that as well.

dotansimha:
> Hi @B2o5T , is there an ETA for releasing v3? I'm currently blocked from that as well.

We'll do it later this week! :)

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@3.0.0`

PR #752, opened 2021-11-04
[breaking change] feat: split `recommended` config to two configs `schema-recommended` and `operations-recommended`

B2o5T:
fix https://github.com/dotansimha/graphql-eslint/issues/552 https://github.com/dotansimha/graphql-eslint/issues/645


packages/plugin/src/configs/schema-all.ts
```diff
@@ -2,37 +2,22 @@
...
     ],
-    '@graphql-eslint/description-style': 'error',
     '@graphql-eslint/input-name': 'error',
-    '@graphql-eslint/match-document-filename': 'error',
-    '@graphql-eslint/no-deprecated': 'error',
-    '@graphql-eslint/no-duplicate-fields': 'error',
-    '@graphql-eslint/no-hashtag-description': 'error',
     '@graphql-eslint/no-scalar-result-type-on-mutation': 'error',
-    '@graphql-eslint/no-unreachable-types': 'error',
     '@graphql-eslint/no-unused-fields': 'error',
```
JustinTRoss, line 18: If the intention is to tailor this set to items that deal with schema only and do not require operations, e.g. for a server environment, consider removing this rule.

B2o5T: the purpose of `all` config is to include all available rules for schema or operations, so this rule is included explicitly

PR #771, opened 2021-11-12
Docs: Enhance basic config instructions.

JustinTRoss:
*Any change needs to be discussed before proceeding. Failure to do so may result in the rejection of the pull request.*

## Description

Enhanced basic configuration instructions in the docs pursuant to @B2o5T's explanation in https://github.com/dotansimha/graphql-eslint/discussions/757

## Type of change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,11 @@ npm install --save-dev @graphql-eslint/eslint-plugin
 
 ### Configuration
 
-To get started, create an override configuration for your ESLint, while applying it to `.graphql` files (do that even if you are declaring your operations in code files):
+#### Setup ESLint override for `.graphql` files.
+
+_This step is necessary even if you are declaring operations and/or schema in code files._
+
+To get started, define an override in your ESLint config to tell ESLint to modify the way it treats `.graphql` files.
 
 ```json
 {
@@ -63,7 +67,9 @@ To get started, create an override configuration for your ESLint, while applying
 }
 ```
 
-If you are using code files to store your GraphQL schema or your GraphQL operations, you can extend the behaviour of ESLint and extract those, by adding **an additional `override`** that does that extraction processes:
+#### Setup ESLint override for code files that define GraphQL schema or operations.
+
+If you are defining GraphQL schema or GraphQL operations in code files, you'll want to define an additional override to extend the functionality of this plugin to the GraphQL schema and operations in those files.
 
 ```json
 {
@@ -83,6 +89,8 @@ If you are using code files to store your GraphQL schema or your GraphQL operati
   ]
 }
 ```
+
+Under the hood, specifying the `@graphql-eslint/graphql` processor for code files will cause `graphql-eslint/graphql` to extract the schema and operations from these files into virtual GraphQL documents with `.graphql` extensions. This will allow the overrides you've defined for `.graphql` files, via `files: ["*.graphql"]`, to get applied to the schema and operations defined in your code files.
 
 #### Extended linting rules with GraphQL Schema
 
--- a/README.md
+++ b/README.md
@@ -80,19 +80,19 @@ To get started, define an override in your ESLint config to tell ESLint to modif
 }
 ```
 
-If your GraphQL definitions are defined only in `.graphql` files and you're only using rules that apply to individual files, you should be good to go :thumbsup:. If you would like to use rules that apply across schema and operation definitions across multiple files, see [here](#using-rules-with-constraints-that-span-the-entire-schema).
+If your GraphQL definitions are defined only in `.graphql` files and you're only using rules that apply to individual files, you should be good to go 👍 . If you would like use a remote schema or use rules that apply across the entire collection of definitions at once, see [here](#using-rules-with-constraints-that-span-the-entire-schema).
 
 #### Tell ESLint to apply this plugin to GraphQL definitions defined in code files.
 
-If you are defining GraphQL schema or GraphQL operations in code files, you'll want to define an additional override to extend the functionality of this plugin to the GraphQL schema and operations in those files.
+If you are defining GraphQL schema or GraphQL operations in code files, you'll want to define an additional override to extend the functionality of this plugin to the schema and operations in those files.
 
 ```diff
 {
   "overrides": [
-+    {
-+      "files": ["*.tsx", "*.ts", "*.jsx", "*.js"],
-+      "processor": "@graphql-eslint/graphql"
-+    },
++   {
++     "files": ["*.tsx", "*.ts", "*.jsx", "*.js"],
++     "processor": "@graphql-eslint/graphql"
++   },
     {
       "files": ["*.graphql"],
       "parser": "@graphql-eslint/eslint-plugin",
@@ -120,7 +120,7 @@ If you are defining GraphQL schema or GraphQL operations in code files, you'll w
 
 Under the hood, specifying the `@graphql-eslint/graphql` processor for code files will cause `graphql-eslint/graphql` to extract the schema and operation definitions from these files into virtual GraphQL documents with `.graphql` extensions. This will allow the overrides you've defined for `.graphql` files, via `files: ["*.graphql"]`, to get applied to the definitions defined in your code files.
 
-#### Using rules with constraints that span the entire schema.
+#### Using a remote schema or rules with constraints that span the entire schema.
 
 Some rules require an understanding of the entire schema at once. For example, [no-unreachable-types](https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/no-unreachable-types.md#no-unreachable-types) checks that all types are reachable by root-level fields.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
       "parser": "@graphql-eslint/eslint-plugin",
       "plugins": ["@graphql-eslint"],
       "rules": {
-        ...
+        "@graphql-eslint/require-description": [
```
JustinTRoss, line 63: Added the most commonly applicable rule I could think of, such that this snippet would be enough to deliver value independently.

Unfortunately, the `require-description` rule config is quite verbose. I know @B2o5T has made some changes for the coming major release which reduce this verbosity. I'm curious if it could be reduced further to accept `["error"]` and just default to the complete set of definitions.

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
+          },
+        ],
       }
     }
   ]
 }
 ```
 
-If you are using code files to store your GraphQL schema or your GraphQL operations, you can extend the behaviour of ESLint and extract those, by adding **an additional `override`** that does that extraction processes:
+If your GraphQL definitions are defined only in `.graphql` files and you're only using rules that apply to individual files, you should be good to go :thumbsup:. If you would like to use rules that apply across schema and operation definitions across multiple files, see [here](#using-rules-with-constraints-that-span-the-entire-schema).
```
JustinTRoss, line 83:
```suggestion
If your GraphQL definitions are defined only in `.graphql` files and you're only using rules that apply to individual files, you should be good to go 👍.. If you would like to use rules that apply to schema and operation definitions across the entire collection of definitions at once, see [here](#using-rules-with-constraints-that-span-the-entire-schema).
```

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
 
-```json
+#### Tell ESLint to apply this plugin to GraphQL definitions defined in code files.
+
+If you are defining GraphQL schema or GraphQL operations in code files, you'll want to define an additional override to extend the functionality of this plugin to the GraphQL schema and operations in those files.
```
JustinTRoss, line 87:
```suggestion
If you are defining GraphQL schema or GraphQL operations in code files, you'll want to define an additional override to extend the functionality of this plugin to the schema and operations in those files.
```

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
+
+```diff
```
JustinTRoss, line 89: Changed formatting to diff to make it clear what's actually changing here.

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
++      "files": ["*.tsx", "*.ts", "*.jsx", "*.js"],
++      "processor": "@graphql-eslint/graphql"
++    },
     {
       "files": ["*.graphql"],
       "parser": "@graphql-eslint/eslint-plugin",
       "plugins": ["@graphql-eslint"],
       "rules": {
-        ...
+        "@graphql-eslint/require-description": [
```
JustinTRoss, line 101: Added the rule here as well to make this a complete valid value-delivering snippet.

See other comments about the specific rule above.

B2o5T:
Amazing, thanks for contributing 😍

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
+          },
+        ],
       }
     }
   ]
 }
 ```
 
-#### Extended linting rules with GraphQL Schema
+Under the hood, specifying the `@graphql-eslint/graphql` processor for code files will cause `graphql-eslint/graphql` to extract the schema and operation definitions from these files into virtual GraphQL documents with `.graphql` extensions. This will allow the overrides you've defined for `.graphql` files, via `files: ["*.graphql"]`, to get applied to the definitions defined in your code files.
```
JustinTRoss, line 121: Explaining what this additional override is doing per discussion https://github.com/dotansimha/graphql-eslint/discussions/757

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
-Linting process can be enriched and extended with GraphQL type information, if you are able to provide your GraphQL schema.
+Some rules require an understanding of the entire schema at once. For example, [no-unreachable-types](https://github.com/dotansimha/graphql-eslint/blob/master/docs/rules/no-unreachable-types.md#no-unreachable-types) checks that all types are reachable by root-level fields.
 
-The parser allow you to specify a json file / graphql files(s) / url / raw string to locate your schema (We are using `graphql-tools` to do that). Just add `parserOptions.schema` to your configuration file:
+To use these rules, you'll need to tell ESLint how to identify the entire set of schema definitions.
 
-```json
+If you are using [`graphql-config`](https://graphql-config.com/), you are good to go. `graphql-eslint` integrates with it automatically and will use it to load your schema!
+
+Alternatively, you can define `parserOptions.schema` in the override in your eslint config.
```
JustinTRoss, line 131:
```suggestion
Alternatively, you can define `parserOptions.schema` in the `*.graphql` override in your ESLint config.
```

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
+
+```diff
 {
   "files": ["*.graphql"],
   "parser": "@graphql-eslint/eslint-plugin",
   "plugins": ["@graphql-eslint"],
-  "parserOptions": {
-    "schema": "./schema.graphql"
-  }
+  "rules": {
```
JustinTRoss, line 140: Adding example rule to make this snippet independently value-delivering. An alternative would be to extend `schema/recommended` functionality, but I don't know if it'll be suitable until after major version release.

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
+```diff
 {
   "overrides": [
-    {
-      "files": ["*.tsx", "*.ts", "*.jsx", "*.js"],
-      "processor": "@graphql-eslint/graphql"
-    },
++    {
++      "files": ["*.tsx", "*.ts", "*.jsx", "*.js"],
++      "processor": "@graphql-eslint/graphql"
```
B2o5T, line 94:
```suggestion
+     "processor": "@graphql-eslint/graphql"
```
Remove one space will be better 
778B1E6A-B0F4-4A57-88E1-322B6A0522F3

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
     }
   ]
 }
 ```
 
-#### Extended linting rules with GraphQL Schema
+Under the hood, specifying the `@graphql-eslint/graphql` processor for code files will cause `graphql-eslint/graphql` to extract the schema and operation definitions from these files into virtual GraphQL documents with `.graphql` extensions. This will allow the overrides you've defined for `.graphql` files, via `files: ["*.graphql"]`, to get applied to the definitions defined in your code files.
 
-If you are using [`graphql-config`](https://graphql-config.com/) - you are good to go. This package integrates with it automatically, and will use it to load your schema!
+#### Using rules with constraints that span the entire schema.
```
JustinTRoss, line 123: We probably need another section that explains configuration of `parserOptions.schema`, which can acknowledge use-case of remote schema. That section can then just be referenced here. I'll add to this PR.

README.md
JustinTRoss: Ahh. Good catch. Will fix.

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
   "files": ["*.graphql"],
   "parser": "@graphql-eslint/eslint-plugin",
   "plugins": ["@graphql-eslint"],
-  "parserOptions": {
-    "schema": "./schema.graphql"
-  }
+  "rules": {
+    "no-unused-types": ["error"]
+  },
++  "parserOptions": {
```
JustinTRoss, line 143: Same here 😅

README.md
```diff
@@ -56,56 +60,95 @@ To get started, create an override configuration for your ESLint, while applying
...
++    "schema": "./schema.graphql"
```
JustinTRoss, line 144: I'll update this to `diff` for operations as well. I don't have bandwidth to evaluate those docs presently though, as I don't have a natural use-case in front of me. I'll circle back there once I'm using it.

README.md
JustinTRoss: Meh. Just changing title to serve both.

dotansimha (approved):
Looks great! 
@B2o5T feel free to merge when you think it's ready :)

B2o5T (approved):

README.md
```diff
@@ -117,13 +160,16 @@ To workaround that, we allow you to provide additional information on your Graph
...
 To provide that, we are using `@graphql-tools` loaders to load your sibling operations and fragments, just specify a glob expression(s) that points to your code/.graphql files:
 
-```json
+```diff
 {
   "files": ["*.graphql"],
   "parser": "@graphql-eslint/eslint-plugin",
   "plugins": ["@graphql-eslint"],
+  "rules": {
+    "unique-operation-name": ["error"]
```
B2o5T, line 169: must be `@graphql-eslint/unique-operation-name`, btw don't worry I'll fix this

JustinTRoss: Good catch! The same probably needs added to line 141.

README.md
B2o5T: I think we can add some simple rule like `no-anonymous-operations` instead.

And about `require-description` rule, I think this rule can't have a default set of definitions - because somebody wants to require description only for `fields`, somebody only for `types`, and somebody only for `ObjectTypeDefinition`.

JustinTRoss: 🤷‍♂️ Whatever defaults are easiest to expect for the user and also safely not violating any best practices. I imagine descriptions to all definitions is a safe default, and then folks can back it off from there or specify their own.

Issue #817, opened 2021-12-04
no-hashtag-description rule applies to operations


dobesv:
**Describe the bug**

It appears that the `@graphql-eslint/no-hashtag-description` rule will apply to operations, even though you cannot add descriptions to operations using a string.

**To Reproduce**

```
# This is a comment that will trigger no-hashtag-description
query Q {
  thisIsAQuery {
    id
  }
}
```

**Expected behavior**

`no-hashtag-description` should only trigger for fields & types.

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@3.1.0` 🎉 thanks @B2o5T

PR #818, opened 2021-12-04
fix: ignore operations and fragments in `no-hashtag-description` rule

B2o5T:
related https://github.com/dotansimha/graphql-eslint/issues/817

### DIFF
--- a/packages/plugin/src/rules/no-hashtag-description.ts
+++ b/packages/plugin/src/rules/no-hashtag-description.ts
@@ -1,4 +1,4 @@
-import { TokenKind } from 'graphql';
+import { Kind, TokenKind } from 'graphql';
 import { GraphQLESLintRule } from '../types';
 import { getLocation } from '../utils';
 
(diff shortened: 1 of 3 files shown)


packages/plugin/src/rules/no-hashtag-description.ts
```diff
@@ -56,16 +55,17 @@ const rule: GraphQLESLintRule = {
     schema: [],
   },
   create(context) {
+    const selector = `${Kind.DOCUMENT}[definitions.0.kind!=/^(${Kind.OPERATION_DEFINITION}|${Kind.FRAGMENT_DEFINITION})$/]`;
```
dotansimha, line 58: does `definitions.0` will effect files with multiple documents? (for example: a document with query and a fragment)

B2o5T: This selector will skip documents where first definition is either operation or fragment

Issue #825, opened 2021-12-07
@graphql-eslint/fields-on-correct-type vscode highlight


tonytoth:
**Describe the bug**
The wrong field is not highlighted the right way in VSCode. As you can see it is working fine in "Problems" tab but the hightlight is not working as expected.

**To Reproduce**
<img width="2077" alt="Screenshot 2021-12-07 at 09 29 05" src="https://user-images.githubusercontent.com/8457868/144985334-32eef274-0480-42d1-b8d0-7340a6174f57.png">

**Expected behavior**
It should highlight the whole line / word.

**Environment:**

- OS: Ubuntu
- `@graphql-eslint/eslint-plugin": "^3.0.1`
- NodeJS: v12.18.3

**Additional context**

B2o5T:
we use `graphql-js` rule that unfortunately report only `start` position of `line` and `column`, I asked this question in their repo https://github.com/graphql/graphql-js/discussions/3415 if it's possible I'll fix this

B2o5T:
Found a workaround to fix this for all `graphql-js` rules, this will be fixed on the next `graphql-eslint` version.

dotansimha:
Fixed in `@graphql-eslint/eslint-plugin@3.2.0`. Thanks @B2o5T ❤️

PR #837, opened 2021-12-11
fix error report location for `graphql-js` rules

B2o5T:
fix https://github.com/dotansimha/graphql-eslint/issues/825

@dotansimha you can explore snapshots to see that now we'll have a correct error report location for `graphql-js` rules 😀.


dotansimha (approved):
PERFECT!!!!

PR #839, opened 2021-12-11
Add suggestions for `naming convention` rule

B2o5T:
https://user-images.githubusercontent.com/7361780/145687520-87fd0914-016b-4674-a06b-b6478a3110c6.mov

### DIFF
--- a/packages/plugin/src/rules/naming-convention.ts
+++ b/packages/plugin/src/rules/naming-convention.ts
@@ -1,6 +1,6 @@
 import { ASTKindToNode, Kind, NameNode } from 'graphql';
 import { GraphQLESLintRule, ValueOf } from '../types';
-import { TYPES_KINDS, getLocation } from '../utils';
+import { TYPES_KINDS, getLocation, convertCase, CaseStyle } from '../utils';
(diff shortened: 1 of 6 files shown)


packages/plugin/src/rules/naming-convention.ts
```diff
@@ -243,69 +244,94 @@ const rule: GraphQLESLintRule<[NamingConventionRuleConfig]> = {
...
+            errorMessage: `not have "${forbiddenPrefix}" prefix`,
+            renameToName: nodeName.replace(new RegExp(`^${forbiddenPrefix}`), ''),
+          };
         }
         const forbiddenSuffix = forbiddenSuffixes?.find(suffix => name.endsWith(suffix));
         if (forbiddenSuffix) {
-          return `not have "${forbiddenSuffix}" suffix`;
-        }
-        if (style && !ALLOWED_STYLES.includes(style)) {
-          return `be in one of the following options: ${ALLOWED_STYLES.join(', ')}`;
```
B2o5T, line 285: don't need because JSON schema already catch this case

PR #912, opened 2022-01-19
feat: add new option `OperationDefinition` in `require-description` rule

B2o5T:
fixes https://github.com/dotansimha/graphql-eslint/discussions/907


packages/plugin/src/rules/require-description.ts
```diff
@@ -27,7 +28,7 @@ const rule: GraphQLESLintRule<[RequireDescriptionRuleConfig]> = {
     docs: {
       category: 'Schema',
       description: 'Enforce descriptions in your type definitions.',
```
dotansimha, line 30: @B2o5T I think it worth mentioning that it's not only for type definitions, but also for operations (using comments syntax instead of description syntax)

B2o5T: mentioned

Issue #838, opened 2021-12-11
Fix `ESTree-AST` converter because currently it's incorrectly convert `loc.end`


B2o5T:
Example: `loc.end` always equal `loc.start`
```js
 {
   token: {
     type: 'Name',
     loc: { start: { line: 4, column: 13 }, end: { line: 4, column: 13 } },
     value: 'veryBad',
     range: [ 40, 47 ]
   }
 }
```

B2o5T:
I think is a bug with `graphql-js` because for `Kind.Field` it's returns incorrect `endToken`
![image](https://user-images.githubusercontent.com/7361780/145682267-725b8fbd-c021-4b29-b305-9c4672192384.png)

PR #939, opened 2022-01-29
fix ESTree parser, convert `loc.column` to 0-based column number

B2o5T:
related https://github.com/dotansimha/graphql-eslint/issues/838

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,7 @@ module.exports = {
   reportUnusedDisableDirectives: true,
   ignorePatterns: ['examples'],
   parser: '@typescript-eslint/parser',
-  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended', 'standard', 'prettier'],
+  extends: ['eslint:recommended', 'standard', 'plugin:@typescript-eslint/recommended', 'prettier'],
   plugins: ['unicorn'],
   rules: {
     'no-empty': 'off',
--- a/packages/plugin/src/estree-parser/utils.ts
+++ b/packages/plugin/src/estree-parser/utils.ts
@@ -9,10 +9,13 @@ import {
   GraphQLNamedType,
   isNonNullType,
   isListType,
+  Token,
+  Source,
+  Lexer,
 } from 'graphql';
-import { Comment } from 'estree';
-import { GraphQLESTreeNode } from './estree-ast';
-import { convertToken } from '../utils';
+import type { Comment } from 'estree';
+import type { AST } from 'eslint';
+import type { GraphQLESTreeNode } from './estree-ast';
 
 export default function keyValMap<T, V>(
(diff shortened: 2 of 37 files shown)


.eslintrc.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   reportUnusedDisableDirectives: true,
   ignorePatterns: ['examples'],
   parser: '@typescript-eslint/parser',
-  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended', 'standard', 'prettier'],
+  extends: ['eslint:recommended', 'standard', 'plugin:@typescript-eslint/recommended', 'prettier'],
```
B2o5T, line 5: `eslint-config-standart` is a config for javascript files (for typescript we must use this https://github.com/standard/eslint-config-standard-with-typescript), due `standard` was placed after `plugin:@typescript-eslint/recommended` there are some false-positive errors because `@typescript-eslint/recommended` turn-out some rules and standard turn-in

packages/plugin/src/estree-parser/utils.ts
```diff
@@ -69,37 +66,19 @@ export function extractCommentsFromAst(loc: Location): Comment[] {
...
 
   while (token !== null) {
-    const { kind, value, line, column, start, end, next } = token;
-    if (kind === TokenKind.COMMENT) {
-      comments.push({
-        type: 'Block',
-        value,
-        loc: {
-          start: { line, column },
-          end: { line, column },
```
B2o5T, line 79: this is a mistake, `end` is always as `start` I fixed it by `column - 1 + (end - start)` where `end` is `range[1]` and start is `range[0]`.
> Note: `-1` is needed because ESLint treat columns as 0 based

packages/plugin/src/rules/input-name.ts
```diff
@@ -108,7 +107,7 @@ const rule: GraphQLESLintRule<[InputNameRuleConfig]> = {
         if (shouldCheckType((node as any).parent.parent)) {
           const name = node.name.value;
           context.report({
-            loc: getLocation(node.loc, name),
+            node: node.name,
```
B2o5T, line 110: because I fixed ESTree parser, now it works as expected

packages/plugin/src/utils.ts
```diff
@@ -84,27 +84,39 @@ function getLexer(source: Source): Lexer {
...
-    tokens.push({
-      type: token.kind as any,
-      loc: {
-        start: {
-          line: token.line,
-          column: token.column,
-        },
-        end: {
-          line: token.line,
-          column: token.column,
```
B2o5T, line 102: this was also a mistake, `end` is always as `start`

dotansimha (approved):

dotansimha:
that's wayyy better, thanks @B2o5T !

PR #975, opened 2022-02-21
fix `no-duplicate-fields`, compare selections with fragment spreads and inline fragments

B2o5T:


dotansimha (approved):

saihaj (commented):
If we think of using fragments all the way down then this can lead to a lot of false positives.

B2o5T:
> If we think of using fragments all the way down then this can lead to a lot of false positives.

An example of a false positive?

packages/plugin/tests/no-duplicate-fields.spec.ts
```diff
@@ -48,7 +48,58 @@ ruleTester.runGraphQLTests('no-duplicate-fields', rule, {
...
+          users {
+            id
+            ...UserFullFields # email #email
+            name #3
+            ...UserFields # email firstName
+            ... on User {
+              id #6
+              name #7
+              email #8
+            }
```
saihaj, line 65: @B2o5T imagine for some nested UI components it is likely you are using `UserFullFields` in one place and `UserFields` in some other place and it can actually be a valid use case but this rule would mark that as error

PR #1019, opened 2022-04-05
Support aliases in require-id-when-available

### LINKED ISSUE #1017: Support selecting IDs with an alias in require-id-when-available rule
In our code base we sometimes select the `id` of a type with an alias as follows:

```gql
someType {
  id: someOtherUniqueField
}
```

In this case the `require-id-when-available` rule will fail, because `id` is in fact not selected directly. However, using such an alias guarantees that the resolved data will contain an `id` field (if this is an unique key is another question). So I think the rule should also support this pattern.

I have some [code](https://github.com/grubersjoe/graphql-eslint/commit/b9080ec089873d91b6b69b75bc915d235ab5d7c1) that checks if an alias for the configured `id` field exists so that the rule will also pass in this case. Would you be interested in a merge request to add this feature?

grubersjoe:
## Description

Selecting the `id` field with an alias should also be valid as discussed in #1017 .

Fixes #1017 

## Type of change

- [x] This change requires a documentation update (I think)

## How Has This Been Tested?

I've added a test. Do we need more tests?

## Questions

I'm not sure if I should bump the version etc. Should I do that? Or will you perform the changes required for a new release?

### DIFF
--- a/.changeset/dirty-rules-sin.md
+++ b/.changeset/dirty-rules-sin.md
@@ -1,0 +1,5 @@
+---
+'@graphql-eslint/eslint-plugin': patch
+---
+
+feat: support select `id` field with an alias in `require-id-when-available` rule
--- a/docs/rules/require-id-when-available.md
+++ b/docs/rules/require-id-when-available.md
@@ -50,6 +50,13 @@ query {
     name
   }
 }
+
+# Selecting `id` with an alias is also valid
+query {
+  user {
+    id: name
+  }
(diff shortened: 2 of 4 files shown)


B2o5T:
> I'm not sure if I should bump the version etc. Should I do that? Or will you perform the changes required for a new release?

I'll take care of it, thanks for your PR

> I've added a test. Do we need more tests?

It's fine

PR #1089, opened 2022-06-17
Add failed test case for selection-set-depth's suggestion feature

tgohn:
After https://github.com/B2o5T/graphql-eslint/pull/971, for
`selection-set-depth` rules, if the depth error occures in a different
file than source file (e.g. in an imported fragment file), JavaScript
exception might thrown because it can not find the token node in source
file.


tgohn:
My apologises,
I created this PR by mistake, was meant to create an issue instead.

PR #1093, opened 2022-06-18
Adds try/catch to the ESLint processor preprocess task

therealgilles:
## Description
Adds try/catch to the ESLint processor preprocess task, to bail and return the original code if an error is thrown by the underlying parser (e.g. a parsing error outside the template strings).

The ESLint crew told me processors are not supposed to throw errors. See discussion here: https://github.com/eslint/eslint/discussions/16015

Fixes [# (issue)](https://github.com/B2o5T/graphql-eslint/issues/1092)

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

Here is a simple repo that reproduces the issue: https://github.com/therealgilles/graphql-config-code-file-ts

## How Has This Been Tested?
I'm not sure how to build the repo and run existing tests. If you have documentation on this, I'll be happy to do it.

**Test Environment**:

- OS: MacOS Ventura 13.0 Beta
- `@graphql-eslint/...`: 3.10.4
- NodeJS: v16.15.0

## Checklist:

- [N/A] I have made corresponding changes to the documentation
- [N/A] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/packages/plugin/src/processor.ts
+++ b/packages/plugin/src/processor.ts
@@ -12,27 +12,32 @@ const blocksMap = new Map<string, Block[]>();
 export const processor: Linter.Processor<Block | string> = {
   supportsAutofix: true,
   preprocess(code, filePath) {
-    if (RELEVANT_KEYWORDS.every(keyword => !code.includes(keyword))) {
+    try {
+      if (RELEVANT_KEYWORDS.every(keyword => !code.includes(keyword))) {
+        return [code];
+      }
+      const extractedDocuments = parseCode({
+        code,
+        filePath,
+        options: {
+          globalGqlIdentifierName: ['gql', 'graphql'],
+          skipIndent: true,
+        },
+      });
+
+      const blocks: Block[] = extractedDocuments.map(item => ({
+        filename: 'document.graphql',
+        text: item.content,
+        lineOffset: item.loc.start.line - 1,
+        offset: item.start + 1,
+      }));
+      blocksMap.set(filePath, blocks);
+
+      return [...blocks, code /* source code must be provided and be last */];
+    } catch {
+      // Bail and return original code if error is thrown.
       return [code];
     }
-    const extractedDocuments = parseCode({
-      code,
-      filePath,
-      options: {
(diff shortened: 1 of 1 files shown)


JounQin:
An error indicate bug, you should never suppress unexpected errors.

therealgilles:
@JounQin The syntax error will be reported by ESLint so it will be caught.

Ideally @babel/parser would be called with the ErrorRecovery flag set to true. But graphql-tag-pluck does not seem to give any control on that. Additionally it's not clear that this would be enough in all parsing error cases.

I think the right thing to do is to suppress the error so that ESLint can proceed and the syntax error can be reported. I would add a warning message when running ESLint in debug mode, so that debugging is easier.

JounQin:
This is similar situation for all processors.

If the syntax is correct, then it's a bug of this plugin.

If the syntax is incorrect, then a clearer crash error should help the user to understand.

Maybe ESLint should support `SyntaxError` with `loc` info from processors. (This is a ESLint proposal, if your description from ESLint team's view is correct.)

therealgilles:
See ESLint discussion here: https://github.com/eslint/eslint/discussions/16015#discussioncomment-3028452

JounQin:
> See ESLint discussion here: [eslint/eslint#16015 (reply in thread)](https://github.com/eslint/eslint/discussions/16015#discussioncomment-3028452)

So if I don't understand incorrectly, @btmills agree to `try/catch` in ESLint's `process` logic instead of handling it by processor itself?

https://github.com/eslint/eslint/discussions/16015#discussioncomment-3028452

therealgilles:
I think it is under consideration. I still think it's bad form for a preprocessor to throw, but not sure how preprocessor would be intended to report errors to ESLint core.

dotansimha:
@B2o5T what do you think?

JounQin:
See https://github.com/eslint/eslint/issues/16104 and my PR https://github.com/eslint/eslint/pull/16105

B2o5T:
Sorry for the long response, closing it here since this will be fixed in ESLint directly

PR #1110, opened 2022-07-10
fix(selection-set-depth): fix `TypeError` when suggestion tried to apply to fragment that located in different file

B2o5T:
fixes https://github.com/B2o5T/graphql-eslint/issues/1090


packages/plugin/tests/selection-set-depth.spec.ts
```diff
@@ -95,5 +95,29 @@ ruleTester.runGraphQLTests<[SelectionSetDepthRuleConfig]>('selection-set-depth',
         }
       `,
     },
+    {
+      name: 'suggestions should not throw error when fragment is located in different file',
+      parserOptions: <ParserOptions>{
```
B2o5T, line 100:
```suggestion
      parserOptions: {
```

PR #1303, opened 2022-12-14
extend require-description lint with operation field definitions

tshedor:
## Description

This change extends the `require-description` lint to require a description of operations. 

```graphql
# Correct
type Query {
  "Fetch users"
  users: [User!]!
}

# Incorrect
type Mutation {
  createUser(user: UserInput!): User!
}
```

The field definition within the operation types is tricky to target. A description isn't required for _every_ `Kind.FIELD_DEFINITION` so the existing feature would've involved excessive documentation. Since good code health could require documentation of usable operations, the lint was extended to target field definitions only within the scope of `Mutation`, `Query`, and `Subscription`.

This adds the boolean option `operationFieldDefinition` to `require-description` (default false).

## Type of change

## How Has This Been Tested?

I added two valid unit tests and two invalid unit tests. They cover when an input is present and when an input is not present.

Both unit tests are within the `require-description.spec.ts` file

**Test Environment**:

- OS: Mac OS Monterrey
- `@graphql-eslint/...`: master?
- NodeJS: 18.3.0

## Checklist:

- [x] I have followed the
- [ ] My changes generate no new warnings (_They add a new ESLint warning that's configurable and default disabled_)
- [ ] Any dependent changes have been merged and published in downstream modules (_N/A_)

### DIFF
--- a/packages/plugin/src/rules/require-description.ts
+++ b/packages/plugin/src/rules/require-description.ts
@@ -20,9 +20,10 @@ type SelectorNode = GraphQLESTreeNode<ValueOf<AllowedKindToNode>>;
 
 export type RequireDescriptionRuleConfig = {
   types?: boolean;
+  operationFieldDefinition?: boolean;
 } & {
-  [key in AllowedKind]?: boolean;
-};
+    [key in AllowedKind]?: boolean;
+  };
 
 function getNodeName(node: SelectorNode) {
   const DisplayNodeNameMap = {
@@ -95,6 +96,16 @@ const rule: GraphQLESLintRule<[RequireDescriptionRuleConfig]> = {
             }
           `,
         },
+        {
+          title: 'Correct',
+          usage: [{ operationFieldDefinition: true }],
+          code: /* GraphQL */ `
+            type Mutation {
+              "Create a new user"
+              createUser: User
+            }
+          `,
+        },
       ],
       configOptions: [
         {
@@ -121,6 +132,10 @@ const rule: GraphQLESLintRule<[RequireDescriptionRuleConfig]> = {
             type: 'boolean',
             description: `Includes:\n\n${TYPES_KINDS.map(kind => `- \`${kind}\``).join('\n')}`,
           },
+          operationFieldDefinition: {
+            type: 'boolean',
+            description: 'Definitions within Query, Mutation, and Subscription'
+          },
           ...Object.fromEntries(
             [...ALLOWED_KINDS].sort().map(kind => {
               let description = `Read more about this kind on [spec.graphql.org](https://spec.graphql.org/October2021/#${kind}).`;
@@ -136,7 +151,7 @@ const rule: GraphQLESLintRule<[RequireDescriptionRuleConfig]> = {
     },
   },
   create(context) {
-    const { types, ...restOptions } = context.options[0] || {};
+    const { types, operationFieldDefinition, ...restOptions } = context.options[0] || {};
 
     const kinds = new Set<string>(types ? TYPES_KINDS : []);
     for (const [kind, isEnabled] of Object.entries(restOptions)) {
@@ -145,6 +160,10 @@ const rule: GraphQLESLintRule<[RequireDescriptionRuleConfig]> = {
       } else {
         kinds.delete(kind);
       }
+    }
+
+    if (operationFieldDefinition) {
+      kinds.add(':matches(ObjectTypeDefinition, ObjectTypeExtension) > FieldDefinition');
     }
 
     const selector = [...kinds].join(',');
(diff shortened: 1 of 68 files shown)


packages/plugin/tests/require-description.spec.ts
```diff
@@ -73,6 +73,24 @@ ruleTester.runGraphQLTests<[RequireDescriptionRuleConfig]>('require-description'
...
+          user(id: ID!): User
+        }
+      `,
+      options: [{ operationFieldDefinition: true }],
+    },
+    {
+      code: /* GraphQL */ `
+        type Query {
+          "Get users"
+          user: [User!]!
```
Victorrent, line 89:
```suggestion
          users: [User!]!
```

Victorrent (approved):

Josiassejod1 (approved):

B2o5T:
Thanks for working on it! I definitely see the benefit of the new option, and will review it soon!

packages/plugin/src/rules/require-description.ts
```diff
@@ -95,6 +96,16 @@ const rule: GraphQLESLintRule<[RequireDescriptionRuleConfig]> = {
...
           `,
         },
+        {
+          title: 'Correct',
+          usage: [{ operationFieldDefinition: true }],
+          code: /* GraphQL */ `
+            type Mutation {
+              "Create a new user"
+              createUser: User
+            }
```
B2o5T, line 106: it will be better to include another type to show that description is only required for the root type field

packages/plugin/src/rules/require-description.ts
```diff
@@ -95,6 +96,16 @@ const rule: GraphQLESLintRule<[RequireDescriptionRuleConfig]> = {
             }
           `,
         },
+        {
+          title: 'Correct',
+          usage: [{ operationFieldDefinition: true }],
```
B2o5T, line 101: I propose to rename this option to `rootField`

packages/plugin/src/rules/require-description.ts
```diff
@@ -147,6 +162,10 @@ const rule: GraphQLESLintRule<[RequireDescriptionRuleConfig]> = {
       }
     }
 
+    if (operationFieldDefinition) {
+      kinds.add(':matches(ObjectTypeDefinition, ObjectTypeExtension)[name.value=/Query|Mutation|Subscription/] > FieldDefinition');
```
B2o5T, line 166: this will not work for root types that are named differently.
example:
```graphql
type MyQuery {
  user: User
}

schema {
  query: MyQuery
}

Comment (B2o5T) [reply packages/plugin/src/rules/require-description.ts:166]: example of fixing https://github.com/B2o5T/graphql-eslint/blob/3e8f71f7ef60dd6e8d5a4605e7ca0bdd2cec2f86/packages/plugin/src/rules/no-root-type.ts#L76

File: packages/plugin/tests/require-description.spec.ts [hunk 5]
```diff
@@ -197,5 +218,25 @@ ruleTester.runGraphQLTests<[RequireDescriptionRuleConfig]>('require-description'
...
+    {
+      code: 'type Query { user(id: String!): User! }',
+      options: [{ operationFieldDefinition: true }],
+      errors: [{ message: 'Description is required for `Query.user`.' }],
+    },
+    {
+      code: 'type Query { users: [User!]! }',
+      options: [{ operationFieldDefinition: true }],
+      errors: [{ message: 'Description is required for `Query.users`.' }],
+    },
```
B2o5T, line 230:
```suggestion
```
one test for `Query` is enough, instead include 1 test with a root type that is named differently than `Query`/`Mutation` or `Subscription`

packages/plugin/src/rules/require-description.ts
tshedor: I add these in, but they're hardcoded to avoid requiring the schema for the whole rule. If you'd rather these not be hardcoded and to match `no-root-type`'s dynamic fetching of the names, I'd be happy to convert this rule so that the schema is exposed to it.

packages/plugin/tests/require-description.spec.ts
tshedor: @B2o5T I added a test case for a root type with [a field as a Correct case](https://github.com/B2o5T/graphql-eslint/pull/1303/commits/6c36d2778c4500ec65a13a0ff4ab8fd0869d73af#diff-a303039c21671d709870c86d53e96a474dff89cd481f292e6fb278ee868bb233R99-R101). Is this what you had in mind? If I misread you I'd be happy to try again

B2o5T (approved):
thanks, will be a part of next release

PR #1310, opened 2022-12-19
Simplify wording in rule docs

FloEdelmann:
*Note: I deliberately ignored the PR description template, because this is a documentation-only change, and the template is not fitting here.*

The previous wording in the `lone-anonymous-operation` rule docs was difficult to understand. I simplified and clarified it.


B2o5T (changes requested):
this sentence is copy-c copy-v from `graphql-js` rule. But I'm okay with this change, thanks!

could you change this description in https://github.com/B2o5T/graphql-eslint/blob/a2cab16538a1ddc326a30669aaf161de846e03d5/packages/plugin/src/rules/graphql-js-validation.ts#L336-L337

and regenerate documentation with following command
https://github.com/B2o5T/graphql-eslint/blob/a2cab16538a1ddc326a30669aaf161de846e03d5/package.json#L10

PR #1317, opened 2022-12-20
Add category icons to rule list in docs

FloEdelmann:
This is a documentation-only change.

When looking through the list of rules, it's good to see at a glance which rules apply to schema documents and which apply to operations.

This PR adds another column with icons for both.


B2o5T (approved):
I like it!

PR #1316, opened 2022-12-20
Create new lone-executable-definition rule

FloEdelmann:
## Description

Adds the suggested rule from #1311 to require one file per query/fragment/mutation/subscription. Fixes #1311.

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

I've created new unit tests for the rule, and also ran `npm test` to be sure that I did not break anything.

**Test Environment**:

- OS: Ubuntu 22.04
- NodeJS: 16.18.0

## Checklist:

- [x] I have followed the
      [CONTRIBUTING](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the
      style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests and linter rules pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

## Further comments

The latest commit (b0b0cc802ea5d66da3d1939416df81b08284d6d2) were automatically generated with `npm run prettier`. I'm happy to revert it if required.

PR #1319, opened 2022-12-20
Follow-up changes to `lone-executable-definition` rule

FloEdelmann:
Thanks @B2o5T for the fixes and refactoring in #1316. I have a few minor follow-up fixes for those.

### DIFF
--- a/.changeset/strange-donuts-warn.md
+++ b/.changeset/strange-donuts-warn.md
@@ -2,4 +2,4 @@
 '@graphql-eslint/eslint-plugin': minor
 ---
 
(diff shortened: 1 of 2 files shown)


B2o5T (changes requested):
Unfortunately, we can't use `OperationTypeNode` since we support graphql 15 and type check throw and error since it's not an enum here, we can drop GraphQL 15 in the next major once ESLint 9 is released

FloEdelmann:
Ah, I see. Since you already fixed the typo in #1320, I'm closing this PR then.

PR #1325, opened 2022-12-21
DRY in `lone-executable-definition` rule

FloEdelmann:
Follow-up to #1316 and #1319. This is a small refactoring to avoid repeating the same GraphQL definition types in the `type` and the schema definition.


B2o5T (approved):
Thanks !

PR #1331, opened 2022-12-22
add new rule `require-type-pattern-with-oneof`

B2o5T:
fixes https://github.com/B2o5T/graphql-eslint/issues/1137

### DIFF
--- a/packages/plugin/tests/require-type-pattern-with-oneof.spec.ts
+++ b/packages/plugin/tests/require-type-pattern-with-oneof.spec.ts
@@ -1,0 +1,55 @@
+import { GraphQLRuleTester } from '../src';
+import { rule } from '../src/rules/require-type-pattern-with-oneof';
+
+const ruleTester = new GraphQLRuleTester();
+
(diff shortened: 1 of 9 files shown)


packages/plugin/tests/require-type-pattern-with-oneof.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const ruleTester = new GraphQLRuleTester();
+
+ruleTester.runGraphQLTests('require-type-pattern-with-oneof', rule, {
+  valid: [
+    /* GraphQL */ `
+      type T @oneOf {
+        ok: Ok
+        error: Error
+      }
```
B2o5T, line 12: should I requires schema and validate that `type T` is used as the result type for some mutation field?

dotansimha: We don't need that at the moment, mainly because `@oneOf` might be used for other things, not just mutations

PR #1334, opened 2022-12-22
add new rule `no-one-place-fragments`

B2o5T:
fixes https://github.com/B2o5T/graphql-eslint/issues/72

> find fragments and variables that used only in one place and suggest to inline them.


B2o5T:
@dotansimha would be great hear your thoughts 😅

PR #1377, opened 2023-01-11
New GraphQL-ESLint website x Nextra

B2o5T:


dotansimha (approved):
This is amazing, thanks @B2o5T 
Do you think we should have an easy state sharing? Like TypeScript playground does? (when you click on Share/Cmd+S, it creates a hash for all the data, set it as a query param, and then you can share the link)
I think it might be useful to help to reproduce issues and give better support to users, without having a heavy reproduction. 

We did the same thing back then on `schemehub`, so here's a quick reference in case you want to add that feature:
* capture cmd+s on Monaco (in your case, it's probably for all editors? or capture it in another way?) https://github.com/the-guild-org/schemehub/blob/3550c7d5993603006431577119492cee352737a1/components/schema-viewer.tsx#L12-L32 
* Calculate hash, replace URL, and copy to clipboard https://github.com/the-guild-org/schemehub/blob/3550c7d5993603006431577119492cee352737a1/pages/index.tsx#L210-L226

Issue #1416, opened 2023-01-25
New rule: Report import expressions without the named fragment


FloEdelmann:
I'd like to suggest a rule that reports [`# import` expressions](https://the-guild.dev/graphql/tools/docs/schema-loading#using-import-expression) which try to import fragments that are not part of that file.

Given file 'foo.graphql' contains a fragment called 'Foo':

### Fail

```graphql
# import Bar from 'foo.graphql'
#        ^^^ File 'foo.graphql' does not contain a 'Bar' fragment.

query MyQuery {
  bar
}
```

### Pass

```graphql
# import Foo from 'foo.graphql'

query MyQuery {
  bar
}
```

```graphql
# import 'foo.graphql'

query MyQuery {
  bar
}
```

B2o5T:
Accepted

Issue #1414, opened 2023-01-25
New rule: Report unused import comment


FloEdelmann:
I'd like to suggest a rule that reports [`# import` comments](https://the-guild.dev/graphql/tools/docs/schema-loading#using-import-expression) that are not used in the current file:

### Fail

```graphql
# import Foo from 'someOtherFile.graphql'
#        ^^^ Fragment 'Foo' not used in this file, import expression can be removed.

query MyQuery {
  bar
}
```

### Pass

```graphql
# import Foo from 'someOtherFile.graphql'

query MyQuery {
  fooField {
    ...Foo
  }
}
```

```graphql
# import 'someOtherFile.graphql'

query MyQuery {
  fooField
}
```

B2o5T:
Accepted

PR #1443, opened 2023-02-07
Add `require-import-fragment` rule

### LINKED ISSUE #1413: New rule: require import expressions for fragments
`graphql-let` and other tools rely on `# import` expressions in GraphQL files to resolve fragments defined in other files: https://github.com/piglovesyou/graphql-let#how-to-import-graphql-from-another-document-especially-graphql-fragment

I'd like to suggest a rule that requires such a comment when a fragment is used:

### Fail

```graphql
query MyQuery {
  fooField {
    ...Foo
#      ^^^ Expected 'Foo' fragment to be imported
  }
}
```

```graphql
# import 'someOtherFile.graphql'

query MyQuery {
  fooField {
    ...Foo
#      ^^^ Expected 'Foo' fragment to be imported
  }
}
```

### Pass

```graphql
# import Foo from 'someOtherFile.graphql'

query MyQuery {
  fooField {
    ...Foo
  }
}
```

```graphql
#import Foo from 'someOtherFile.graphql'

query MyQuery {
  fooField {
    ...Foo
  }
}
```

FloEdelmann:
## Description

Fixes #1413

## Type of change

## How Has This Been Tested?

Via unit tests (see changed files).

**Test Environment**:

- OS: Ubuntu 22.04
- NodeJS: 16.19.0

## Checklist:

- [x] I have followed the


packages/plugin/src/rules/require-import-fragment.ts
```diff
@@ -0,0 +1,116 @@
...
+import { GraphQLESLintRule } from '../types.js';
+
+const RULE_ID = 'require-import-fragment';
+const SUGGESTION_ID = 'add-import-expression';
+
+export const rule: GraphQLESLintRule = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      category: 'Operations',
```
B2o5T, line 12: schema types could be imported too with import comments

packages/plugin/src/rules/require-import-fragment.ts
```diff
@@ -0,0 +1,116 @@
...
+      description: 'Require fragments to be imported via an import expression.',
```
B2o5T, line 13:
```suggestion
      description: 'Require documents to be imported via an import expression `# import …`.',
```
maybe?

packages/plugin/src/rules/require-import-fragment.ts
```diff
@@ -0,0 +1,116 @@
...
+    hasSuggestions: true,
+    messages: {
+      [RULE_ID]: "Expected '{{name}}' fragment to be imported.",
+      [SUGGESTION_ID]: "Add import expression for '{{name}}'",
+    },
+    schema: [],
+  },
+  create(context) {
+    const knownFragmentNames = new Set<string>();
+    const fragmentSpreadNodes = new Set<GraphQLESTreeNode<FragmentSpreadNode>>();
```
B2o5T, line 62:
```suggestion
    const fragmentSpreadNameNodes = new Set<GraphQLESTreeNode<NameNode>>();
```
we can simplify to push just fragment NameNode

packages/plugin/src/rules/require-import-fragment.ts
```diff
@@ -0,0 +1,116 @@
...
+    const comments = context.getSourceCode().getAllComments();
+
+    function checkFragmentSpreadNode(fragmentSpreadNode: GraphQLESTreeNode<FragmentSpreadNode>) {
```
B2o5T, line 65:
```suggestion
    function checkFragmentSpreadNode(nameNode: GraphQLESTreeNode<NameNode>) {
```

packages/plugin/src/rules/require-import-fragment.ts
```diff
@@ -0,0 +1,116 @@
...
+      const fragmentName = fragmentSpreadNode.name.value;
```
B2o5T, line 66:
```suggestion
      const fragmentName = nameNode.value;
```

packages/plugin/src/rules/require-import-fragment.ts
```diff
@@ -0,0 +1,116 @@
...
+        if (
+          comment.type === 'Line' &&
+          comment.value.trim().startsWith(`import ${fragmentName} from `)
+        ) {
+          return;
+        }
+      }
+
+      context.report({
+        node: fragmentSpreadNode.name,
```
B2o5T, line 82:
```suggestion
        node: nameNode,
```

packages/plugin/src/rules/require-import-fragment.ts
```diff
@@ -0,0 +1,116 @@
...
+              );
+            },
+          },
+        ],
+      });
+    }
+
+    return {
+      FragmentSpread(fragmentSpreadNode) {
+        fragmentSpreadNodes.add(fragmentSpreadNode);
```
B2o5T, line 104:
```suggestion
        fragmentSpreadNodes.add(fragmentSpreadNode.name);
```
I guess I nothing missed

packages/plugin/src/rules/require-import-fragment.ts
```diff
@@ -0,0 +1,116 @@
...
+      const fragmentName = fragmentSpreadNode.name.value;
+
+      if (knownFragmentNames.has(fragmentName)) {
+        return;
+      }
+
+      for (const comment of comments) {
+        if (
+          comment.type === 'Line' &&
+          comment.value.trim().startsWith(`import ${fragmentName} from `)
```
B2o5T, line 75:
```suggestion
          comment.value.trimStart().startsWith(`import ${fragmentName} from `)
```

packages/plugin/tests/require-import-fragment.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        # import 'foo.graphql'
+
+        query MyQuery {
+          fooField {
+            ...Foo
+          }
+        }
+      `,
+      errors: [{ message: "Expected 'Foo' fragment to be imported." }],
+    },
```
B2o5T, line 59: I guess it's a valid case? and the fragment will be still imported?

FloEdelmann: I thought that I have seen this syntax supported somewhere, but I can't find any reference to it. So I'll adjust the test description to `should report fragments when there are only invalid import expressions`

B2o5T: it's a valid import https://github.com/ardatan/graphql-tools/blob/37c71441751801e6a7bfca6966ffe7f83e7992eb/packages/import/tests/documents/import-documents.spec.ts#L9-L29

and fixture

https://github.com/ardatan/graphql-tools/blob/37c71441751801e6a7bfca6966ffe7f83e7992eb/packages/import/tests/documents/import-test/default/a.graphql#L1-L7

B2o5T:
also, what will happen if you import `not-foo.graphql` and this file does not contain `Foo` fragment

```graphql
# import Foo from 'not-foo.graphql'

        query MyQuery {
          fooField {
            ...Foo
          }
        }
```

packages/plugin/tests/require-import-fragment.spec.ts
FloEdelmann: Ah, I see. I guess these are our options:

1. Keep current behavior, i.e. treat default imports `# import 'foo.graphql'` as non-existent and require specific imports `# import Foo from 'foo.graphql'`.
  This would allow this rule to be enabled without information about other operation files.
  Importing `Foo` from `not-foo.graphql` would not be treated as an error by this rule, but instead be reported by a separate rule (that requires operation file information): #1416
2. Treat all fragment spreads as valid as soon as there is a default import `# import 'foo.graphql'`. This would work without operation file information, too. But which rule would then check whether `Foo` is really in `foo.graphql`?
  Maybe #1414? But then that one would need operation file information, while it currently doesn't.
3. Require operation file information in this rule and explicitly check whether default imports contain the specified fragment.
4. Add an option to this rule that specifies how to treat default imports:
  * `defaultImports: 'ignore'` = option 1
  * `defaultImports: 'check' = option 3, would then require operation file information
  * `defaultImports: 'forbid' → disallow default imports altogether. But that should probably be a separate rule.

Or some other option or a combination of the above.

B2o5T: > ignore default imports # import 'foo.graphql' and require specific imports # import Foo from 'foo.graphql'.

no, we should validate the both

> This would allow this rule to be enabled without information about other operation files.

this rule should require operations files to be smart

> Add an option to this rule that specifies how to treat default imports:

I guess we don't need options


I will help you today-tomorrow, and will push directly in your PR some logic, so you could reuse for other rules

B2o5T: @FloEdelmann check what I made ;)

packages/plugin/src/rules/require-import-fragment.ts
```diff
@@ -100,11 +126,11 @@ export const rule: GraphQLESLintRule = {
...
 
     return {
-      FragmentSpread(fragmentSpreadNode) {
-        fragmentSpreadNameNodes.add(fragmentSpreadNode.name);
+      'FragmentSpread .name'(node: GraphQLESTreeNode<NameNode>) {
+        fragmentSpreadNameNodes.add(node);
       },
-      FragmentDefinition(fragmentDefinitionNode) {
-        knownFragmentNames.add(fragmentDefinitionNode.name.value);
+      'FragmentDefinition .name'(node: GraphQLESTreeNode<NameNode>) {
```
FloEdelmann: Why `.name` and not just `Name`?

B2o5T: it is because `FragmentDefinition Name` will match

```js
// some abstract GraphQL AST Node
const node = {
  type: 'FragmentDefinition',
  name: { // ✅ matched 1
    type: 'Name'
    // ...
  },
  foo: { // ✅ matched 2
    type: 'Name',
    bar: { // ✅ matched 3
      type: 'Name' 
      // ...
    }
    // ...
  }
  // ...
}
```
2. `FragmentDefinition .name` will match only `node.name`
3. `FragmentDefinition > Name` will match only `node.name` and `node.foo` but not `not.foo.bar`

So for more safety, it should be `FragmentDefinition > .name`, I remember I got some issue due too `wide` selector, so it's better to be explicit while selecting nodes

B2o5T (approved):
thanks for working on it! In case you would work on other rules you could extract some pieces of code for parsing`#import` comments!

PR #1540, opened 2023-03-25
Add `requiredPrefixes` and `requiredSuffixes` to `naming-convention`

louisscruz:
## Description

https://github.com/B2o5T/graphql-eslint/issues/1521

This adds two new configuration options to `naming-convention`:

* `requiredPrefixes` – if supplied, the selected field name must have one of the supplied prefixes
* `requiredSuffixes` – if supplied, the selected field name must have one of the supplied suffixes

Fixes #[1521](https://github.com/B2o5T/graphql-eslint/issues/1521)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as
      expected)
- [ ] This change requires a documentation update

Adding links to sandbox or providing screenshots can help us understand more about this PR and take
action on it as appropriate

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can
reproduce. Please also list any relevant details for your test configuration

- [x] Added relevant tests in `naming-convention.spec.ts`

## Checklist:

- [x] I have followed the

### DIFF
--- a/packages/plugin/src/rules/naming-convention.ts
+++ b/packages/plugin/src/rules/naming-convention.ts
@@ -112,24 +112,12 @@ const createRequiredPrefixNameSuggestions = (
   requiredPrefixes: string[],
   name: string,
 ): string[] => {
-  switch (style) {
-    case undefined: {
-      return requiredPrefixes.map(requiredPrefix => `${requiredPrefix}${name}`);
-    }
-    case 'camelCase':
-    case 'PascalCase':
-    case 'UPPER_CASE': {
-      return requiredPrefixes.map(
-        requiredPrefix => requiredPrefix + name.charAt(0).toUpperCase() + name.slice(1),
-      );
-    }
-    case 'snake_case':
-    case 'kebab-case': {
-      const joiningCharacter = style === 'kebab-case' ? '-' : '_';
+  if (style === undefined)
+    return requiredPrefixes.map(requiredPrefix => `${requiredPrefix}${name}`);
 
-      return requiredPrefixes.map(requiredPrefix => `${requiredPrefix}${joiningCharacter}${name}`);
-    }
-  }
+  // The underscore here is used to delineate a word, regardless of style,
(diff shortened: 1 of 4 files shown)


packages/plugin/src/rules/naming-convention.ts
```diff
@@ -105,6 +107,98 @@ const schema = {
   },
 } as const;
 
+const createRequiredPrefixNameSuggestions = (
```
B2o5T, line 110: I am surprised with changed lines for such as small option. Plz do not overcomplicate this rule, use convertCase function from utils.ts

louisscruz: > Plz do not overcomplicate this rule, use convertCase function from utils.ts

I see one spot in this function where `convertCase` can be used, but it'll only reduce the number of characters added in this PR by a handful. For the bulk of the logic `convertCase` is insufficient for what I think needs to be done here.

As a user, I'd expect that given configuration like this:

```ts
  'FieldDefinition[gqlType.name.value=Boolean]': {
    style: 'camelCase',
    requiredPrefixes: ['is', 'has'],
  },
```

and schema like this:

```gql
type Thing {
  enabled: Boolean
}
```

that I would be given a suggestion like `Rename to "isEnabled"`. I would be disappointed if it showed `Rename to "isenabled"`.

```ts
convertCase(`${requiredPrefix}${name}`); // -> "isenabled"
```

Unless `convertCase` is adjusted to somehow parse English into words and use that information to convert casing, I don't think that can be used to remove most lines here.

I can take a pass to see if the number of lines introduced in this PR can be reduced though.

louisscruz: Ah, ok, there did end up being a way to lean on `convertCase`. Pushed!

B2o5T (approved):

PR #1665, opened 2023-05-23
fix `TypeError: Cannot read properties of undefined (reading 'kind')` for `require-nullable-result-in-root` rule

B2o5T:
fixes https://github.com/B2o5T/graphql-eslint/issues/1663


nishtahir (approved):
LGTM. Thanks for the quick turnaround

PR #2049, opened 2023-12-01
[new rule] validate subgraph according federation 2

dimaMachina:
Part of https://github.com/dimaMachina/graphql-eslint/issues/1824

I am new in the Federation spec and feel like validation should be split into 2 rules

- validation of supergraph
- validation of subgraph (current PR)

### DIFF
--- a/packages/plugin/package.json
+++ b/packages/plugin/package.json
@@ -39,8 +39,14 @@
     "typecheck": "tsc --noEmit"
   },
   "peerDependencies": {
+    "@apollo/federation-internals": "^2",
     "eslint": ">=8.44.0",
     "graphql": "^16"
+  },
+  "peerDependenciesMeta": {
+    "@apollo/federation-internals": {
+      "optional": true
+    }
   },
   "dependencies": {
     "@graphql-tools/code-file-loader": "^8.0.0",
@@ -53,6 +59,7 @@
     "lodash.lowercase": "^4.3.0"
   },
   "devDependencies": {
+    "@apollo/federation-internals": "^2.6.1",
     "@theguild/eslint-rule-tester": "workspace:*",
     "@types/debug": "4.1.12",
     "@types/eslint": "8.44.8",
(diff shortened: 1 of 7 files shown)


packages/plugin/package.json
```diff
@@ -39,9 +39,15 @@
...
   "peerDependencies": {
+    "@apollo/federation-internals": "^2",
     "eslint": ">=8.44.0",
     "graphql": "^16"
   },
+  "peerDependenciesMeta": {
+    "@apollo/federation-internals": {
+      "optional": true
+    }
+  },
```
dimaMachina, line 50: I think people should explicitly install this dependency if they want to validate federation schema, there is no need to add to dependencies

packages/plugin/src/rules/federation-subgraph.ts
```diff
@@ -0,0 +1,62 @@
+import {
+  asFed2SubgraphDocument,
+  buildSchemaFromAST,
+  FederationBlueprint,
+  Subgraph,
+} from '@apollo/federation-internals';
+import { GraphQLESLintRule } from '../types.js';
+
+const RULE_ID = 'federation-subgraph';
```
dimaMachina, line 9: any suggestions to rule name?

packages/plugin/package.json
```diff
@@ -39,9 +39,15 @@
...
   "dependencies": {
     "@graphql-tools/code-file-loader": "^8.0.0",
     "@graphql-tools/graphql-tag-pluck": "^8.0.0",
```
dimaMachina, line 53: maybe we should add `@graphql-tools/graphql-tag-pluck` also to `peerDependencies` in v4, since it's a big dependency that installs a lot of things (not everybody lint code files)

cgriego:
What's needed to move this forward?

packages/plugin/src/rules/federation-subgraph.ts
vrochamaia: `apollo-federation-subgraph` would be more precise?

dotansimha (approved):

PR #2281, opened 2024-05-03
Add ignore config to `no-unused-fields` rule

### LINKED ISSUE #2280: Add ignore list to `no-unused-fields` rule
**Is your feature request related to a problem? Please describe.**

The `no-unused-fields` rule is a great way to avoid bloat in my schema, but I'm willing to accept that bloat in some cases. For example, I'm not worried if I don't request all the relay pagination fields for every connection exposed in my schema. Currently, though, that's a dealbreaker preventing me from using this rule.

**Describe the solution you'd like**

Any configuration option to provide a list of fields that are allowed to be unused. I'm not particularly wedded to any format for that configuration, but ~I'll provide a PR shortly that showcases one option.~ see PR [here](https://github.com/dimaMachina/graphql-eslint/pull/2281) with one possible option.

**Additional context**

n/a

maciesielka:
## Description

Changes here add a `ignoredFieldSelectors` configuration to the `no-unused-fields` rule such that some fields are allowed to be unused.

Fixes #2280

## Type of change

## Checklist:

- [x] I have followed the


packages/plugin/src/rules/no-unused-fields.ts
```diff
@@ -1,10 +1,39 @@
...
 const RULE_ID = 'no-unused-fields';
 
+const schema = {
+  type: 'array',
+  maxItems: 1,
+  items: {
+    type: 'object',
+    additionalProperties: false,
+    properties: {
+      ignoredFieldSelectors: {
```
maciesielka, line 17: A few other names I considered here were:
- `allowedUnusedFields`
- `allowedUnusedFieldSelectors`
- `ignoredFields`

I think `ignoredFieldSelectors` avoids some of the oxymoronic qualities of `allowedUnused*` and provided useful context that this configuration uses ESLint selectors, but am open to other options.

packages/plugin/src/rules/no-unused-fields.ts
```diff
@@ -1,10 +1,39 @@
...
+        type: 'array',
+        uniqueItems: true,
+        minItems: 1,
+        description: [
+          'Fields that will be ignored and are allowed to be unused.',
+          '',
+          '> These fields are defined by ESLint [`selectors`](https://eslint.org/docs/developer-guide/selectors). Paste or drop code into the editor in [ASTExplorer](https://astexplorer.net) and inspect the generated AST to compose your selector.',
```
maciesielka, line 24: The [`naming-convention` rule](https://github.com/dimaMachina/graphql-eslint/blob/master/packages/plugin/src/rules/naming-convention.ts) already makes use of ESLint selectors within configurations and it felt good to follow suit here. While this adds some clunkiness to the configuration (allowing an individual field looks like `"FieldDefinition[name.value=field]"` instead of "field"), there are benefits in that allowing fields with wildcards, based on their parent object, etc. are implicitly supported.

packages/plugin/src/rules/no-unused-fields.ts
dimaMachina: I prefer `ignoredFieldSelectors` too

packages/plugin/src/rules/no-unused-fields.ts
dimaMachina: I think it's fine to use selectors here instead of plain strings because we have the flexibility to choose parent name and parent pattern, see

```js
[
  '[parent.name.value=PageInfo][name.value=endCursor]',
  '[parent.name.value=PageInfo][name.value=startCursor]',
  '[parent.name.value=PageInfo][name.value=hasNextPage]',
  '[parent.name.value=PageInfo][name.value=hasPreviousPage]',
  '[parent.name.value=/Edge$/][name.value=cursor]',
  '[parent.name.value=/Connection$/][name.value=pageInfo]',
]
```

we can specify end's with `Edge` and `Connection`, @maciesielka maybe we should update the default options to include it there?

packages/plugin/src/rules/no-unused-fields/index.test.ts
```diff
@@ -95,6 +98,17 @@ ruleTester.run('no-unused-fields', rule, {
         },
       },
     },
+    {
+      name: 'should allow fields if they are ignored',
```
dimaMachina, line 102:
```suggestion
      name: 'should do not report unused fields for relay',
```

dimaMachina (approved):
Sorry for taking time merging it, great work!

PR #2711, opened 2024-11-17
Reenable running `require-id-when-available` on `FragmentDefinition`

dimaMachina:
fixes https://github.com/dimaMachina/graphql-eslint/issues/1737


Stephen2:
ohhhh this is wonderful, very much appreciated 🙏

PR #2735, opened 2024-11-20
Fix report lint issues not on first char of file for `.vue` and support ESLint fixes and suggestions

dimaMachina:
fixes https://github.com/dimaMachina/graphql-eslint/issues/2103
cc @bmulholland


examples/vue-code-file/eslint.config.js
```diff
@@ -1,18 +1,31 @@
...
+    // `eslint-plugin-vue` will set a default processor for `.vue` files
+    // we use `eslint-merge-processors` to extend it
+    processor: mergeProcessors([
+      pluginVue.processors['.vue'],
+      processorVueBlocks({
+        blocks: {
+          styles: false,
+          customBlocks: true,
+          script: true,
+          template: false,
```
dimaMachina, line 25: I think we can't declare graphql documents in template? can be removed? @bmulholland
```suggestion
```

examples/vue-code-file/eslint.config.js
```diff
@@ -1,18 +1,31 @@
...
-      parser: vueParser,
-    },
+    // `eslint-plugin-vue` will set a default processor for `.vue` files
+    // we use `eslint-merge-processors` to extend it
+    processor: mergeProcessors([
+      pluginVue.processors['.vue'],
+      processorVueBlocks({
+        blocks: {
+          styles: false,
+          customBlocks: true,
```
dimaMachina, line 23: should I remove processing custom blocks? @bmulholland

examples/vue-code-file/eslint.config.js
```diff
@@ -1,18 +1,31 @@
...
+          script: true,
```
dimaMachina, line 24: also what about this `scriptSetup`

https://github.com/antfu/eslint-processor-vue-blocks/blob/9826f22e4180d95adda2edd1585e65d248e6577b/src/index.ts#L69-L74

examples/vue-code-file/eslint.config.js
```diff
@@ -1,18 +1,31 @@
...
 import graphqlPlugin from '@graphql-eslint/eslint-plugin';
 
 export default [
   {
-    files: ['**/*.js', '**/*.vue'],
+    files: ['**/*.js'],
     processor: graphqlPlugin.processor,
     rules: js.configs.recommended.rules,
   },
+  ...pluginVue.configs['flat/recommended'],
```
dimaMachina, line 13: i used this recommended config to show more real world example

bmulholland:
Oh wow this is really exciting! I'll take a look :)

examples/vue-code-file/eslint.config.js
bmulholland: I didn't even know you could do that! They sound cool... and like something that should be out of scope for now.

They're custom, so you'd want to pretty careful with the scenario. Best for someone to have the need and share what it is in more detail. That would probably be a very long tail thing though... honestly probably can always ignore.

examples/vue-code-file/eslint.config.js
bmulholland: Script setup is essentially the new standard, best practice way to write Vue components. Almost all my code is in that, or migrating to it. Should certainly be supported.

examples/vue-code-file/eslint.config.js
bmulholland: I suspect you _can_. But people _shouldn't_. It certainly wouldn't be the Vue way, or a best practice. I say don't support it.

examples/vue-code-file/eslint.config.js
```diff
@@ -1,18 +1,31 @@
...
   {
     files: ['**/*.vue'],
-    languageOptions: {
-      parser: vueParser,
-    },
+    // `eslint-plugin-vue` will set a default processor for `.vue` files
+    // we use `eslint-merge-processors` to extend it
```
bmulholland, line 17: FWIW, I don't understand this comment. What is the merge doing, and why?

dimaMachina: I copied it from https://github.com/antfu/eslint-processor-vue-blocks#usage

examples/vue-code-file/eslint.config.js
dimaMachina: ok let's remove this from example

examples/vue-code-file/test.vue
```diff
@@ -2,8 +2,6 @@
   <span>test</span>
 </template>
 <script>
-/* eslint-disable no-unused-vars */
```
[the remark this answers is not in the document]
bmulholland: Yeah it wouldn't be the first time I see a lint disabling a rule and also disabling the unused-directive check hah.

You could easily just use the variables, though. I'll post code above.

examples/vue-code-file/test.vue
```diff
@@ -2,8 +2,6 @@
   <span>test</span>
```
bmulholland, line 2:
```suggestion
	<!-- use the variables to satisfy the no-unused-vars lint -->
  <span>{{ GET_USER }}</span>
  <span>{{ GET_ANOTHER_USER }}</span>
```

examples/vue-code-file/eslint.config.js
bmulholland: Ah! That context helps. How about something like:
> Vue still needs to be parsed by the Vue parser for normal ESLinting. But GraphQL's lint needs to lint only the JS/TS part, so extract those as blocks using `eslint-processor-vue-blocks`. This turns the script parts of Vue SFCs into virtual JS/TS blocks inside ESLint. ESLint can then parse the JS/TS to find GraphQL parts. And finally, graphql-eslint can lint the resulting GraphQL.

examples/vue-code-file/test.vue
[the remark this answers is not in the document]
bmulholland: Did that work?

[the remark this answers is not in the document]
dimaMachina: nope, because it's upstream issue from merge-processors https://github.com/antfu/eslint-processor-vue-blocks/issues/8#issuecomment-2489562965

dimaMachina:
> Nice find on that package!
> 
> If you think this is ready, I can try it out on my project. I always struggle with trying unpackaged JS libraries (spoiled by Ruby, I suppose). Could I just clone this branch and use `pnpm link` on the cloned dir? Is there an easier/better way to try it out?

graphql-eslint needs a new alpha release to try it out, I will do it in few minutes with [bug fix](https://github.com/dimaMachina/graphql-eslint/issues/2724)

you will need just update your project according new vue example

examples/vue-code-file/eslint.config.js
dimaMachina: missed comment, added in https://github.com/dimaMachina/graphql-eslint/commit/e59c13fe169eb3691a5dda6e0291997ad8e3bb60 @bmulholland

PR #2743, opened 2024-11-21
fix error from `no-deprecated` rule `4.0.0-alpha.13: node.typeInfo(...).inputType.getFields is not a function`

dimaMachina:
fixes https://github.com/dimaMachina/graphql-eslint/issues/2724


packages/plugin/src/rules/no-deprecated/index.ts
```diff
@@ -132,14 +132,16 @@ export const rule: GraphQLESLintRule<[], true> = {
...
-        const typeInfo = node.typeInfo();
-        // @ts-expect-error -- fixme
-        const fields = typeInfo.inputType!.getFields();
-        for (const field of node.fields) {
-          const fieldName = field.name.value;
-          const reason = fields[fieldName].deprecationReason;
-          if (reason) {
-            report(field, reason);
+        const { inputType } = node.typeInfo();
+        if (!inputType) return;
```
dimaMachina, line 136:
```suggestion
```

PR #2752, opened 2024-11-23
fix - Config types don't satisfy FlatConfig.Config

dimaMachina:
fixes https://github.com/dimaMachina/graphql-eslint/issues/2442


alecmev:
Thanks! 👍

PR #2765, opened 2024-11-26
fix bug report repo example link

dimaMachina:
fixes https://github.com/dimaMachina/graphql-eslint/issues/2341

### DIFF
--- a/examples/graphql-config/package.json
+++ b/examples/graphql-config/package.json
@@ -13,7 +13,7 @@
   },
   "devDependencies": {
     "@eslint/js": "9.15.0",
-    "@graphql-eslint/eslint-plugin": "workspace:*",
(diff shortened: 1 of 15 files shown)


examples/graphql-config/package.json
```diff
@@ -13,7 +13,7 @@
   },
   "devDependencies": {
     "@eslint/js": "9.15.0",
-    "@graphql-eslint/eslint-plugin": "workspace:*",
+    "@graphql-eslint/eslint-plugin": "4.0.0-alpha.16",
```
bmulholland, line 16: Was this intentional? You'll have to maintain the version now.

dimaMachina: yeah this is intentional, otherwise, we need to modify manually version in Stackblitz, renovate will create a PRs with new versions when they will out

PR #2818, opened 2024-12-05
Link to proper /docs and remove Hive-specific links

hasparus:
## Description

Currently, the Documentation link in the navmenu is going to wrong documentation. Adding `developerMenu` in layout explicitly fixes it.

I made the `developerMenu` prop required in next major of `@theguild/components` to avoid errors like this.

There's still an issue of the <kbd>Sign in</kbd> button being present, and it was decided the libraries will just link to /docs there (it's fixed in `@theguild/components` main branch), so we'll have to bump the version when the Tailwind build issue is fixed.

image

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

image

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can
reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Environment**:

- OS:
- `@graphql-eslint/...`:
- NodeJS:

## Checklist:

### DIFF
--- a/website/app/layout.tsx
+++ b/website/app/layout.tsx
@@ -1,5 +1,5 @@
 import { ComponentPropsWithoutRef, FC, ReactNode } from 'react';
-import { Banner, PRODUCTS } from '@theguild/components';
+import { Banner, GitHubIcon, PaperIcon, PencilIcon, PRODUCTS, RightCornerIcon, TargetIcon } from '@theguild/components';
 import { getDefaultMetadata, GuildLayout } from '@theguild/components/server';
 import '@theguild/components/style.css';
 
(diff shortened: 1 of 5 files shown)


website/app/layout.tsx
```diff
@@ -47,6 +47,19 @@ const RootLayout: FC<{
           { children: 'Rules', href: '/rules' },
           { children: 'Playground', href: '/play' },
         ],
+        developerMenu: [
+          {
+            href: '/docs',
+            icon: PaperIcon,
```
dimaMachina, line 53: did you meant ? you cannot pass functions from server to client components
```suggestion
            icon: <PaperIcon />,
```

hasparus: You're right, but we'll have to update Components first.

hasparus: Done

